PR #6, opened 2018-12-29
Make changes to allow usage in other wallet-related libraries

itswisdomagain:
The changes made so far are not breaking changes and it's expected that the mobilewallet library build for Android or iOS will continue to work as before.

This is a WIP PR. Updates will continue to be added as I implement more features to make this library broader and usable outside its current mobile binary usages.


itswisdomagain:
The changes in this PR will be added to master in smaller, successive PRs beginning with #20.

PR #13, opened 2019-02-02
adrmngr: fix dir and add log

JoeGruffins:
This is not what was causing 0 peers, just something I found along the way.  The addrmngr directory was pointing to a non-existing directory.  It now receives the intended directory.  Also added addrmngr output to logs for future debugging.


JoeGruffins:
passing travis now

PR #31, opened 2019-05-27
Move sync estimation calculation variables to syncData

C-ollins:
Moving sync estimation calculations variables to syncData fixes a bug that causes estimated sync listener to report wrong values when added during sync.


dcrlibwallet.go
```diff
@@ -96,12 +96,33 @@ func newLibWallet(walletDataDir, walletDbDriver string, activeNet *netparams.Par
 		go shutdownListener()
 	}
 
+	// Init Sync Data
```
itswisdomagain: Probably best to do this when sync starts instead of here - `spvSync` and `rpcSync` functions.

sync.go
```diff
@@ -16,11 +16,34 @@ import (
...
-	cancelSync            context.CancelFunc
-	syncProgressListeners []SyncProgressListener
-	rescanning            bool
+	mu                             sync.Mutex
+	rpcClient                      *chain.RPCClient
+	cancelSync                     context.CancelFunc
+	syncProgressListeners          []SyncProgressListener
+	estimatedSyncProgressListeners []EstimatedSyncProgressListener
+	rescanning                     bool
+	syncing                        bool
```
itswisdomagain: Can move these new properties (L25-46) to an `activeSyncData` struct and then embed the new struct here.

sync.go
```diff
@@ -35,31 +58,22 @@ func (lw *LibWallet) AddSyncProgressListener(syncProgressListener SyncProgressLi
...
+	lw.estimatedSyncProgressListeners = append(lw.estimatedSyncProgressListeners, syncProgressListener)
 }
 
 func (lw *LibWallet) SyncInactiveForPeriod(totalInactiveSeconds int64) {
-	for _, syncProgressListener := range lw.syncProgressListeners {
-		if syncProgressEstimator, ok := syncProgressListener.(*SyncProgressEstimator); ok {
-			syncProgressEstimator.DiscardPeriodsOfInactivity(totalInactiveSeconds)
-		}
+
+	lw.totalInactiveSeconds += totalInactiveSeconds
```
itswisdomagain: `lw.syncData.totalInactiveSeconds` reads better.

sync.go
```diff
@@ -35,31 +58,22 @@ func (lw *LibWallet) AddSyncProgressListener(syncProgressListener SyncProgressLi
...
 	}
 }
 
-func (lw *LibWallet) ResetSyncProgressListeners() {
-	for _, syncProgressListener := range lw.syncProgressListeners {
-		if syncProgressEstimator, ok := syncProgressListener.(*SyncProgressEstimator); ok {
-			syncProgressEstimator.Reset()
-		}
-	}
+func (lw *LibWallet) ResetSyncProgressListener() {
```
itswisdomagain: - This can become `initSyncData`, with same code in lines 99-117 of `dcrlibwallet.go`
- Function should probably not be exported to ensure it is only called internally from `spvSync` and `rpcSync`

sync.go
```diff
@@ -113,7 +127,9 @@ func (lw *LibWallet) SpvSync(peerAddresses string) error {
 
 	// syncer.Run uses a wait group to block the thread until sync completes or an error occurs
 	go func() {
+		lw.syncing = true
 		err := syncer.Run(ctx)
+		lw.syncing = false
```
itswisdomagain: I think I'd find it easier to deduce from the following snippet that `lw.syncing` is turned off when the go routine is exited.
```
lw.syncing = true
defer func() {
	lw.syncing = false
}()

err := syncer.Run(ctx)
```

Also need to set `lw.syncing = false` in following callbacks:
- sync canceled
- sync error
- sync completed

syncnotification.go
```diff
@@ -34,99 +31,382 @@ func (lw *LibWallet) spvSyncNotificationCallbacks(loadedWallet *wallet.Wallet) *
...
-			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnDiscoveredAddresses(SyncStateStart)
+
+			lw.totalFetchedHeadersCount = 0
+
+			if lw.showLogs && lw.syncing {
+				log.Info("Fetch headers completed.")
 			}
 		},
+		DiscoverAddressesStarted: lw.discoverAddressesStarted,
```
itswisdomagain: Might be cleaner to maintain this pattern for **all** callback functions, create a separate `lw.callBackHandler` function to host the method body instead of placing the method body here directly.

sync.go
```diff
@@ -30,35 +58,57 @@ const (
...
+		headersFetchProgress:     headersFetchProgress,
+		addressDiscoveryProgress: addressDiscoveryProgress,
+		headersRescanProgress:    headersRescanProgress,
+
+		beginFetchTimeStamp:     -1,
+		headersFetchTimeSpent:   -1,
+		totalDiscoveryTimeSpent: -1,
+	}
+
+	lw.syncData.activeSyncData = activeSyncData
```
itswisdomagain: Can assign `lw.syncData.activeSyncData` directly in line 78.

sync.go
```diff
@@ -30,35 +58,57 @@ const (
...
 
-func (lw *LibWallet) AddEstimatedSyncProgressListener(syncProgressListener EstimatedSyncProgressListener, logEstimatedProgress bool) {
-	syncProgressEstimator := SetupSyncProgressEstimator(
-		lw.activeNet.Params.Name,
-		logEstimatedProgress,
-		lw.GetBestBlock,
-		lw.GetBestBlockTimeStamp,
-		syncProgressListener,
-	)
+func (lw *LibWallet) AddSyncProgressListener(syncProgressListener SyncProgressListener) {
```
itswisdomagain: Also add a `RemoveSyncProgressListener` method. I used a map (dictionary) on dcrios to enable removing listeners whenever necessary. You pass a string `identifier` when registering and deregistering a listener.

sync.go
```diff
@@ -255,36 +317,40 @@ func (lw *LibWallet) RescanBlocks() error {
...
 				log.Error(p.Err)
-
 				return
 			}
-			totalHeight += p.ScannedThrough
+			totalHeightRescanned += p.ScannedThrough
 			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnRescan(p.ScannedThrough, SyncStateProgress)
+				report := &HeadersRescanProgressReport{
+					CurrentRescanHeight: p.ScannedThrough,
```
itswisdomagain: `CurrentRescanHeight` should be `totalHeightRescanned`, `p.ScannedThrough` represents the number of headers scanned in this loop, `CurrentRescanHeight` should be the total number of headers scanned to this point.

sync.go
```diff
@@ -255,36 +317,40 @@ func (lw *LibWallet) RescanBlocks() error {
...
-			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnRescan(totalHeight, SyncStateFinish)
+			select {
+			case <-ctx.Done():
+				log.Info("Rescan cancelled through context")
+				return
+			default:
+				continue
 			}
 		}
```
itswisdomagain: After the for loop completes, add code here to send final rescan progress notification (to be on the safe side). The current rescan height of this last notification would be equal to the total headers to scan value to signify to the app that rescan has completed.

sync.go
```diff
@@ -255,36 +317,40 @@ func (lw *LibWallet) RescanBlocks() error {
...
-			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnRescan(totalHeight, SyncStateProgress)
-			}
-		default:
-			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnRescan(totalHeight, SyncStateFinish)
+			select {
+			case <-ctx.Done():
+				log.Info("Rescan cancelled through context")
+				return
```
itswisdomagain: Errors such as these (in this case, the error is `ctx.Err()`) and the `p.Err` at the start of each loop cycle are currently ignored. Should find a way to notify the app of such errors so the interface(s) are updated properly. I've created an issue for this, to be worked on later: https://github.com/raedahgroup/dcrlibwallet/issues/35

We'd probably be better off introducing a separate `progress listener` for rescan, that includes error and completion callbacks along with the progress callback.

syncnotification.go
```diff
@@ -34,99 +30,397 @@ func (lw *LibWallet) spvSyncNotificationCallbacks(loadedWallet *wallet.Wallet) *
 		RescanProgress:               generalNotifications.RescanProgress,
 		RescanFinished:               generalNotifications.RescanFinished,
 		PeerDisconnected: func(peerCount int32, addr string) {
-			for _, syncResponse := range lw.syncProgressListeners {
-				syncResponse.OnPeerDisconnected(peerCount)
-			}
+			lw.handlePeerCountUpdate(peerCount)
```
itswisdomagain: Please re-order the `PeerDisconnected` and `PeerConnected` callbacks to make them easier to scan. Should be the first set of callbacks listed in this `*spv.Notifications` object.

This `spvSyncNotificationCallbacks` calls `lw.generalSyncNotificationCallbacks()` hence that function (`lw.generalSyncNotificationCallbacks`) is immediately below `spvSyncNotificationCallbacks`.

And immediately below `lw.generalSyncNotificationCallbacks` is `lw.handlePeerCountUpdate`, which is why I'm recommending that the `PeerDisconnected` and `PeerConnected` callbacks go to the top since they are the ones that call `lw.handlePeerCountUpdate()`

Arranging functions this way makes it easier to guess where a function could be in a long file just by looking at the hierarchy of function calls.

syncnotification.go
```diff
@@ -34,99 +30,397 @@ func (lw *LibWallet) spvSyncNotificationCallbacks(loadedWallet *wallet.Wallet) *
...
-		DiscoverAddressesStarted: func() {
-			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnDiscoveredAddresses(SyncStateStart)
-			}
-		},
-		DiscoverAddressesFinished: func() {
-			for _, syncProgressListener := range lw.syncProgressListeners {
-				syncProgressListener.OnDiscoveredAddresses(SyncStateFinish)
-			}
+		Synced:                       lw.synced,
```
itswisdomagain: Also re-arrange these functions in this `*chain.Notifications` object in the order that we expect they'd be called. I think moving  only `Synced` to the bottom (after `RescanFinished`) would achieve that.

Also, re-arrange the equivalent `lw.---` functions to match the order they appear here. `lw. fetchHeadersStarted` should appear first before `lw.fetchHeadersProgress` and so on.

The `lw.handlePeerCountUpdate` function will appear before `lw. fetchHeadersStarted` since it is called first in the scheme of things.

The last set of functions would be `lw.notifySyncCanceled`, `lw.notifySyncError` and `lw.synced`

PR #32, opened 2019-05-28
Add delete wallet function

C-ollins:
This pull request adds a delete function to dcrlibwallet package to enable users delete wallet without errors.

### DIFF
--- a/dcrlibwallet.go
+++ b/dcrlibwallet.go
@@ -161,12 +161,12 @@ func (lw *LibWallet) DeleteWallet(privatePassphrase []byte) error {
 	}()
 
 	wallet, loaded := lw.walletLoader.LoadedWallet()
-	if !loaded{
+	if !loaded {
 		return errors.New(ErrWalletNotLoaded)
 	}
 
 	err := wallet.Unlock(privatePassphrase, nil)
-	if err != nil{
+	if err != nil {
 		return translateError(err)
 	}
 	wallet.Lock()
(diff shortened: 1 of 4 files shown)


dcrlibwallet.go
```diff
@@ -152,6 +154,32 @@ func (lw *LibWallet) Shutdown(exit bool) {
...
 
+func (lw *LibWallet) DeleteWallet(privatePassphrase []byte) error {
+
+	defer func() {
+		for i := range privatePassphrase {
+			privatePassphrase[i] = 0
+		}
+	}()
+
+	lw.CancelSync()
```
itswisdomagain: Is this needed here? `lw.Shutdown` called in line 178 below also calls `lw.CancelSync()`.

sync.go
```diff
@@ -307,6 +307,10 @@ func (lw *LibWallet) CancelSync() {
 
 	lw.walletLoader.SetNetworkBackend(nil)
 	loadedWallet.SetNetworkBackend(nil)
+
+	log.Info("Waiting to lose all peers")
+	peersWG.Wait()
```
itswisdomagain: `CancelSync` can be called at other times without a need to wait for all peers to be lost. Maybe add a boolean `losePeers` parameter to control this function.

syncnotification.go
```diff
@@ -14,13 +15,17 @@ const (
 	SyncStateFinish   = "finish"
 )
 
+var peersWG sync.WaitGroup
```
itswisdomagain: Let's add this to `syncData` instead of making it a global variable

sync.go
```diff
@@ -307,6 +307,14 @@ func (lw *LibWallet) CancelSync() {
 
 	lw.walletLoader.SetNetworkBackend(nil)
 	loadedWallet.SetNetworkBackend(nil)
+
+	// It's important to wait lose all peers when canceling sync
```
itswisdomagain:
```suggestion
	// It's important to wait to lose all peers when canceling sync
	// if the wallet database would be closed after canceling sync
```

PR #34, opened 2019-05-29
correct headers fetch progress estimation code

itswisdomagain:
Issues corrected:
- Formula for calculating stage 1 time remaining and headers fetch rate.
- Using wallet `bestBlockTime` for estimating total headers to fetch. There's an occasional difference between `lastHeaderTime` and wallet `bestBlockTime` which would likely have been responsible for negative block behind reporting on the mobile apps.

Also corrected:
- Accounting for lost time at the end of stage 1, before the start of stage 2.


itswisdomagain:
The work done here is incorporated in #31 which has been merged into master. Closing this.

PR #38, opened 2019-06-01
Use correct endHeight for IndexTransactions log

C-ollins:
The log displayed when indexing transaction displays the best block height as the transaction height. This fixes it by displaying the block height gotten from the currently indexed transactions block header as the transaction height.

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -34,6 +34,7 @@ func (lw *LibWallet) IndexTransactions(beginHeight int32, endHeight int32, after
 	ctx, _ := contextWithShutdownCancel(context.Background())
 
 	var totalIndex int32
+	var txEndHeight uint32
 	rangeFn := func(block *wallet.Block) (bool, error) {
 		for _, transaction := range block.Transactions {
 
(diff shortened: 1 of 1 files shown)


transactions.go
```diff
@@ -60,13 +60,13 @@ func (lw *LibWallet) IndexTransactions(beginHeight int32, endHeight int32, after
 		}
 
 		if block.Header != nil {
-			err := lw.txDB.Set(BucketTxInfo, KeyEndBlock, &endHeight)
+			err := lw.txDB.Set(BucketTxInfo, KeyEndBlock, &block.Header.Height)
```
itswisdomagain: The final log message (in line 110 below) will also benefit from this correction.

C-ollins: Final log should display current best block height since the wallet scanned through those blocks and did not find related transactions.

itswisdomagain: What if there was an error during the scan, and the wallet did not scan through to the best block?

transactions.go
```diff
@@ -60,13 +60,13 @@ func (lw *LibWallet) IndexTransactions(beginHeight int32, endHeight int32, after
...
 			if err != nil {
 				log.Errorf("Set tx index end block height error: ", err)
 				return false, err
 			}
 
-			log.Infof("Transaction index caught up to %d", endHeight)
+			log.Infof("Transaction index caught up to %d", block.Header.Height)
```
itswisdomagain: `Index saved for transactions in block %d`

PR #41, opened 2019-06-03
Sync estimate calculation improvements

itswisdomagain:
- Correct total time remaining calculation during stage 3 (headers rescan)
- Adjust ETA calculation for stage 1 (headers fetch progress). See https://github.com/raedahgroup/dcrlibwallet/issues/43#issuecomment-498949399


C-ollins:
There's a bug that occurs that makes the fetch header remaining time appear as "1m3s" when it's at 100% which is obviously an incorrect estimate. The bug has been traced down to the adjustments and this value was recorded when fetch header progress was at 100%
`First Estimate: 0.000610s, Adjusted Estimate: 63.000610s`

itswisdomagain:
> There's a bug that occurs that makes the fetch header remaining time appear as "1m3s" when it's at 100% which is obviously an incorrect estimate. The bug has been traced down to the adjustments and this value was recorded when fetch header progress was at 100%
> `First Estimate: 0.000610s, Adjusted Estimate: 63.000610s`

Fixed in https://github.com/raedahgroup/dcrlibwallet/pull/41/commits/7cbb7b3a5a83d76e76045a6b4d65ffa4f4075526

PR #47, opened 2019-06-06
rescan: broadcast sync cancelled when rescan is cancelled

C-ollins:
This helps the app reset the display back to normal view

### DIFF
--- a/sync.go
+++ b/sync.go
@@ -345,6 +345,10 @@ func (lw *LibWallet) CancelSync(losePeers bool) {
 	}
 }
 
+func (lw *LibWallet) IsSynced() bool {
+	return !lw.syncData.syncing && lw.connectedPeers != 0
+}
+
 func (lw *LibWallet) IsSyncing() bool {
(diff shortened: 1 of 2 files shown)


syncnotification.go
```diff
@@ -463,6 +465,7 @@ func (lw *LibWallet) notifySyncCanceled() {
 
 func (lw *LibWallet) synced(synced bool) {
 	lw.syncing = false
+	lw.syncData.synced = false
```
itswisdomagain: This should be true.

sync.go
```diff
@@ -419,13 +424,6 @@ func (lw *LibWallet) RescanBlocks() error {
 	return nil
 }
 
-func (lw *LibWallet) CancelRescan() {
-	if lw.syncData.cancelRescan != nil {
```
itswisdomagain: We can as well do away with the `lw.syncData.cancelRescan` variable. It's of no further use with this change.

PR #48, opened 2019-06-10
transactions: add function to query transactions with filter

### LINKED ISSUE #28: Make transaction directions constant
Dcrlibwallet has 3 transaction directions
`SENT` - 0  
`RECEIVED` - 1  
`TRANSFERED` - 2  
These should be declared as constants so it can be accessible outside dcrlibwallet.

C-ollins:
This change makes it easier to query transactions based on type or direction. This also removes sorting of transactions from the frontend and makes it fully based in dcrlibwallet.
Closes #28

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -382,7 +382,8 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
 	return string(result), nil
 }
 
-func (lw *LibWallet) queryTransactions(query storm.Query) (string, error) {
+func (lw *LibWallet) FilterTransactions(txFilter int32) (string, error) {
+	query := lw.prepareTxFilter(txFilter)
 	var transactions []Transaction
 
 	err := query.Find(&transactions)
@@ -398,7 +399,8 @@ func (lw *LibWallet) queryTransactions(query storm.Query) (string, error) {
 	return string(jsonEncodedTransactions), nil
 }
 
-func (lw *LibWallet) countTransactions(query storm.Query) (int, error) {
+func (lw *LibWallet) CountTransactions(txFilter int32) (int, error) {
+	query := lw.prepareTxFilter(txFilter)
 
 	count, err := query.Count(&Transaction{})
 	if err != nil {
@@ -406,16 +408,6 @@ func (lw *LibWallet) countTransactions(query storm.Query) (int, error) {
 	}
 
 	return count, nil
-}
-
-func (lw *LibWallet) FilterTransactions(txFilter int32) (string, error) {
-	query := lw.parseTxFilter(txFilter)
-	return lw.queryTransactions(query)
-}
-
-func (lw *LibWallet) CountTransactions(txFilter int32) (int, error) {
-	query := lw.parseTxFilter(txFilter)
-	return lw.countTransactions(query)
 }
 
 func (lw *LibWallet) DetermineTxFilter(txHash []byte) (int32, error) {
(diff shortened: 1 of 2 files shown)


transactions.go
```diff
@@ -361,6 +382,104 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+	case TxDirectionReceived:
+		return TxFilterReceived, nil
+	default:
+		return TxFilterTransferred, nil
+	}
+}
+
+// - Helper Functions
+
+func (lw *LibWallet) parseTxFilter(txFilter int32) storm.Query {
```
itswisdomagain: - method doesn't actually "parse" some data, can be renamed to say `makeQueryForTxFilter` or `prepareTxQuery`
- Instead of declaring `var query storm.Query` (line 444 below), you can make use a named return variable in this method declaration.

transactions.go
```diff
@@ -361,6 +382,104 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+
+	var query storm.Query
+	switch txFilter {
+	case TxFilterSent:
+		query = lw.txDB.Select(
+			q.Eq("Type", TxTypeRegular),
+			q.Eq("Direction", TxDirectionSent),
```
itswisdomagain: Is it possible for a tx to have direction sent but not type regular?

transactions.go
```diff
@@ -361,6 +382,104 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+	query := lw.parseTxFilter(txFilter)
+	return lw.queryTransactions(query)
+}
+
+func (lw *LibWallet) CountTransactions(txFilter int32) (int, error) {
+	query := lw.parseTxFilter(txFilter)
+	return lw.countTransactions(query)
+}
+
+func (lw *LibWallet) DetermineTxFilter(txHash []byte) (int32, error) {
```
itswisdomagain: What's a use case for this method?

transactions.go
```diff
@@ -361,6 +382,104 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+
+	count, err := query.Count(&Transaction{})
+	if err != nil {
+		return -1, err
+	}
+
+	return count, nil
+}
+
+func (lw *LibWallet) FilterTransactions(txFilter int32) (string, error) {
```
itswisdomagain: I'm thinking `GetTransactions` instead of `FilterTransactions`

transactions.go
```diff
@@ -361,6 +382,104 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+	query := lw.parseTxFilter(txFilter)
+	return lw.queryTransactions(query)
+}
```
itswisdomagain: There don't seem to be any other call to `lw.queryTransactions`, can probably implement the function here directly.

transactions.go
```diff
@@ -361,6 +382,104 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+
+func (lw *LibWallet) CountTransactions(txFilter int32) (int, error) {
+	query := lw.parseTxFilter(txFilter)
+	return lw.countTransactions(query)
```
itswisdomagain: There don't seem to be any other call to `lw. countTransactions`, can probably implement the function here directly.

transactions.go
C-ollins: Check https://github.com/decred/dcrandroid/pull/378

transactions.go
C-ollins: There's already a `GetTransactions` function

transactions.go
```diff
@@ -408,16 +410,6 @@ func (lw *LibWallet) countTransactions(query storm.Query) (int, error) {
...
-	query := lw.parseTxFilter(txFilter)
-	return lw.queryTransactions(query)
-}
-
-func (lw *LibWallet) CountTransactions(txFilter int32) (int, error) {
-	query := lw.parseTxFilter(txFilter)
-	return lw.countTransactions(query)
-}
-
 func (lw *LibWallet) DetermineTxFilter(txHash []byte) (int32, error) {
```
itswisdomagain: The properties needed to determine a tx filter are `tx.Type` and `tx.Direction`. From the use case in decred/dcrandroid#378, those information already exist in the context, no need to pass a txHash and reconstruct the tx just to get those info. You can modify this function to take `tx.Type` and `tx.Direction` or probably just do the determination of tx filter on dcrandroid directly.

transactions.go
```diff
@@ -361,6 +382,91 @@ func (lw *LibWallet) DecodeTransaction(txHash []byte) (string, error) {
...
+	if err != nil {
+		return -1, err
+	}
+
+	return count, nil
+}
+
+func (lw *LibWallet) DetermineTxFilter(txType string, txDirection int32) int32 {
+
+	if txType != TxTypeRegular {
```
itswisdomagain: These nested if statements can be simplified thus:
```
if txType == TxTypeCoinBase {
	return TxFilterCoinBase
}
if txType != TxTypeRegular {
	return TxFilterStaking
}
```

Also, the empty line at the start of the function can be done without.

PR #52, opened 2019-06-28
shutdown: use one goroutine to listen for shutdown signal

C-ollins:
The previous implementation created a goroutine for each context created with shutdown cancel. This implementation uses one goroutine and stores the cancel functions in a slice which is then looped through and invoked once shutdown signal is received.
This change also fixes a bug that causes the code to be stuck when writing to the shutdown channel and there are zero subscribers on the channel.

### DIFF
--- a/dcrlibwallet.go
+++ b/dcrlibwallet.go
@@ -100,6 +100,8 @@ func newLibWallet(walletDataDir, walletDbDriver string, activeNet *netparams.Par
 		syncData:      syncData,
 	}
 
+	listenForShutdown()
+
 	return lw, nil
 }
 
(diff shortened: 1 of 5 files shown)


dcrlibwallet.go
```diff
@@ -100,6 +100,8 @@ func newLibWallet(walletDataDir, walletDbDriver string, activeNet *netparams.Par
 		syncData:      syncData,
 	}
 
+	listenForShutdown()
```
itswisdomagain: Let's tie the shutdown listener to the `lw` instance to ensure proper deallocation of variables when the `lw` instance dies. This also means adding the `shuttingDown` channel and `cancelFuncs` slice to the `LibWallet` struct.

PR #61, opened 2019-08-27
Key-value settings feature

Kifen:
Implemented a key-value settings feature in a newly created settings.go file. The key-value feature was implemented using github.com/asdine/storm


itswisdomagain:
Re-done in #64.

PR #65, opened 2019-09-05
Refactor send function to support sending to multiple recipients simultaneously

Kifen:
Refactor txauthor.go to enable sending to multiple recipients simultaneously while ensuring compatibility with gomobile. This is tested with https://github.com/raedahgroup/dcrios/pull/521.


txauthor.go
```diff
@@ -184,3 +185,135 @@ func (lw *LibWallet) PublishUnminedTransactions() error {
...
+		log.Error(err)
+		return nil, err
+	}
+
+	txHash, err := lw.wallet.PublishTransaction(&msgTx, serializedTransaction.Bytes(), n)
+	if err != nil {
+		return nil, translateError(err)
+	}
+	return txHash[:], nil
+}
```
itswisdomagain: It's a good idea to end files with an extra line space at the bottom

Kifen: Fixed

PR #66, opened 2019-09-05
New method UnspentOutputs

Kifen:
Added function "UnspentOutputs" which was adapted from a similar function on branch dcrlibwallet-wip

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -78,15 +78,14 @@ func (lw *LibWallet) SpendableForAccount(account int32, requiredConfirmations in
 	return int64(bals.Spendable), nil
 }
 
-func (lw *LibWallet) UnspentOutputs(account uint32, requiredConfirmations int32, targetAmount int64) ([]*UnspentOutput, error) {
+func (lw *LibWallet) UnspentOutputs(account uint32, requiredConfirmations int32) ([]*UnspentOutput, error) {
 	policy := wallet.OutputSelectionPolicy{
 		Account:               account,
 		RequiredConfirmations: requiredConfirmations,
 	}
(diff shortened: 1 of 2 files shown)


accounts.go
```diff
@@ -2,6 +2,9 @@ package dcrlibwallet
 
 import (
 	"encoding/json"
+	"fmt"
+	"github.com/decred/dcrd/dcrutil"
```
itswisdomagain: Re-arrange your imports.

Kifen: Imports re-arranged

accounts.go
```diff
@@ -75,6 +78,44 @@ func (lw *LibWallet) SpendableForAccount(account int32, requiredConfirmations in
 	return int64(bals.Spendable), nil
 }
 
+func (lw *LibWallet) UnspentOutputs(account uint32, requiredConfirmations int32, targetAmount int64) ([]*UnspentOutput, error) {
```
itswisdomagain: Take off this `targetAmount` parameter. Use an hardcoded value of 0 in line 86.

accounts.go
```diff
@@ -75,6 +78,44 @@ func (lw *LibWallet) SpendableForAccount(account int32, requiredConfirmations in
...
+	policy := wallet.OutputSelectionPolicy{
+		Account:               account,
+		RequiredConfirmations: requiredConfirmations,
+	}
+	inputDetail, err := lw.wallet.SelectInputs(dcrutil.Amount(targetAmount), policy)
+	// Do not return errors to caller when there was insufficient spendable
+	// outputs available for the target amount.
+	if err != nil && !errors.Is(errors.InsufficientBalance, err) {
```
itswisdomagain: This should become `if err != nil {`. Also take off the comments in lines 87 to 88.

PR #70, opened 2019-09-17
cleanup: init logrotator before returning errors, add utility methods

itswisdomagain:
godcr uses `dcrlibwallet.RegisterLogger` to add a logger to the logrotator, attempting to use that logger to log errors prior to the initialization of the logger by dcrlibwallet causes a panic.

Also,
- create wallet db directory before trying to initialize config db because failing to do so causes an error when attempting to init config db
- Add wordlist and password strength (shannon entropy) functions


itswisdomagain:
Will leave this open to do more testing with godcr so all needed adjustments can be made in this PR before getting it merged.

PR #72, opened 2019-11-12
AccountIterator

bgptr:
GoMobile does not export `Accounts.Acc` property. This PR resolves this with an iterator. 
This feature is used in PR https://github.com/raedahgroup/dcrios/pull/533

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -49,10 +49,17 @@ func (lw *LibWallet) GetAccountsRaw(requiredConfirmations int32) (*Accounts, err
 	}, nil
 }
 
-func (accounts *Accounts) Next() *Account {
-	if accounts.Iter.CurrentIndex < len(accounts.Acc) {
-		account := accounts.Acc[accounts.Iter.CurrentIndex]
-		accounts.Iter.CurrentIndex++
+func (accounts *Accounts) GetAccountIterator() *AccountIterator {
+	return &AccountIterator{
+		CurrentIndex: 0,
+		Acc:          accounts.Acc,
+	}
+}
+
+func (iter *AccountIterator) Next() *Account {
+	if iter.CurrentIndex < len(iter.Acc) {
+		account := iter.Acc[iter.CurrentIndex]
+		iter.CurrentIndex++
 		return account
 	}
 	return nil
--- a/accounts.go
+++ b/accounts.go
@@ -49,6 +49,15 @@ func (lw *LibWallet) GetAccountsRaw(requiredConfirmations int32) (*Accounts, err
 	}, nil
 }
 
+func (accounts *Accounts) Next() *Account {
+	if accounts.Iter.CurrentIndex < len(accounts.Acc) {
+		account := accounts.Acc[accounts.Iter.CurrentIndex]
+		accounts.Iter.CurrentIndex++
+		return account
+	}
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


accounts.go
```diff
@@ -49,6 +49,15 @@ func (lw *LibWallet) GetAccountsRaw(requiredConfirmations int32) (*Accounts, err
 	}, nil
 }
 
+func (accounts *Accounts) Next() *Account {
+	if accounts.Iter.CurrentIndex < len(accounts.Acc) {
```
C-ollins: Can find where `accounts.Iter` was initialized. Do this code run without errors?

bgptr: Yes, it runs. 
I use it like this:

`func walletAccounts(confirmations: Int32) -> [DcrlibwalletAccount] {
        var accounts = [DcrlibwalletAccount]()
        do {
            let accountsRaw = try self.getAccountsRaw(confirmations)
            accountsRaw.iter = DcrlibwalletAccountIterator()
            
            while let account = accountsRaw.next() {
                accounts.append(account)
            }
            
            return accounts
        } catch let error {
            print("Error fetching wallet accounts: \(error.localizedDescription)")
            return accounts
        }
}`

bgptr: Commited a nicer implementation: cb25842

accounts.go
```diff
@@ -49,6 +49,22 @@ func (lw *LibWallet) GetAccountsRaw(requiredConfirmations int32) (*Accounts, err
 	}, nil
 }
 
+func (accounts *Accounts) GetAccountIterator() *AccountIterator {
```
itswisdomagain: The following reads better as `accounts.Iterator` rather than `accounts.GetAccountIterator`:
```suggestion
func (accounts *Accounts) Iterator() *AccountIterator {
```

accounts.go
```diff
@@ -49,6 +49,22 @@ func (lw *LibWallet) GetAccountsRaw(requiredConfirmations int32) (*Accounts, err
...
+	return &AccountIterator{
+		CurrentIndex: 0,
+		Acc:          accounts.Acc,
+	}
+}
+
+func (iter *AccountIterator) Next() *Account {
```
itswisdomagain:
```suggestion
func (accountsInterator *AccountIterator) Next() *Account {
```

types.go
```diff
@@ -40,6 +40,11 @@ type Account struct {
 	ImportedKeyCount int32
 }
 
+type AccountIterator struct {
```
itswisdomagain:
```suggestion
type AccountsIterator struct {
```
Also, the properties in this struct do not need to be public.

types.go
```diff
@@ -40,6 +40,11 @@ type Account struct {
...
+	CurrentIndex int
+	Acc          []*Account
```
itswisdomagain:
```suggestion
	accounts          []*Account
```

go.mod
```diff
@@ -40,3 +40,5 @@ require (
 	golang.org/x/sys v0.0.0-20190422165155-953cdadca894 // indirect
 	google.golang.org/appengine v1.5.0 // indirect
 )
+
+go 1.13
```
itswisdomagain: Please undo the changes to this file.

bgptr: done. sorry!

PR #78, opened 2019-12-22
Reindex transactions after rescan and add rescan progress listener

### LINKED ISSUE #71: Reindex transactions after block rescan
Transactions database should be cleared and reindex after every rescan. This was causing rescan to not be effective when trying to recover the correct state, and required users to reseed instead to see the correct balance and transaction history.

C-ollins:
This change adds a listener interface that broadcasts updates to the frontend when the wallet is rescanning. Transactions saved to the database are cleared after a successful rescan and an index operation is started from block height 0 to the latest block.
Closes #71

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -1,0 +1,116 @@
+package dcrlibwallet
+
+import (
+	"math"
+	"time"
+
+	"github.com/decred/dcrwallet/errors"
+	w "github.com/decred/dcrwallet/wallet/v3"
+)
+
+func (mw *MultiWallet) RescanBlocks(walletID int) error {
+
+	wallet := mw.WalletWithID(walletID)
+	if wallet == nil {
+		return errors.E(ErrNotExist)
+	}
+
+	netBackend, err := wallet.internal.NetworkBackend()
+	if err != nil {
+		return errors.E(ErrNotConnected)
+	}
+
+	if mw.IsRescanning() || !mw.IsSynced() {
+		return errors.E(ErrInvalid)
+	}
+
+	go func() {
+		defer func() {
+			mw.syncData.mu.Lock()
+			mw.syncData.rescanning = false
+			mw.syncData.mu.Unlock()
+		}()
+
+		mw.syncData.mu.Lock()
+		mw.syncData.rescanning = true
+		mw.syncData.mu.Unlock()
+
+		if mw.blocksRescanProgressListener != nil {
+			mw.blocksRescanProgressListener.OnBlocksRescanStarted()
+		}
+
+		ctx, cancel := wallet.shutdownContextWithCancel()
+		mw.syncData.cancelRescan = cancel
+
+		progress := make(chan w.RescanProgress, 1)
+		go wallet.internal.RescanProgressFromHeight(ctx, netBackend, 0, progress)
+
+		rescanStartTime := time.Now().Unix()
+
+		for p := range progress {
+			if p.Err != nil {
+				log.Error(p.Err)
+				if mw.blocksRescanProgressListener != nil {
+					mw.blocksRescanProgressListener.OnBlocksRescanEnded(p.Err)
+				}
+				return
+			}
+
+			rescanProgressReport := &HeadersRescanProgressReport{
+				CurrentRescanHeight: p.ScannedThrough,
+				TotalHeadersToScan:  wallet.GetBestBlock(),
+				WalletID:            walletID,
+			}
+
+			elapsedRescanTime := time.Now().Unix() - rescanStartTime
+			rescanRate := float64(p.ScannedThrough) / float64(rescanProgressReport.TotalHeadersToScan)
+
+			rescanProgressReport.RescanProgress = int32(math.Round(rescanRate * 100))
+			estimatedTotalRescanTime := int64(math.Round(float64(elapsedRescanTime) / rescanRate))
+			rescanProgressReport.RescanTimeRemaining = estimatedTotalRescanTime - elapsedRescanTime
+
+			rescanProgressReport.GeneralSyncProgress = &GeneralSyncProgress{
+				TotalSyncProgress:         rescanProgressReport.RescanProgress,
+				TotalTimeRemainingSeconds: rescanProgressReport.RescanTimeRemaining,
+			}
+
+			if mw.blocksRescanProgressListener != nil {
+				mw.blocksRescanProgressListener.OnBlocksRescanProgress(rescanProgressReport)
+			}
+
+			select {
+			case <-ctx.Done():
+				log.Info("Rescan canceled through context")
+				return
+			default:
+				continue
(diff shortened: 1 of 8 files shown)


rescan.go
```diff
@@ -0,0 +1,116 @@
...
+			rescanRate := float64(p.ScannedThrough) / float64(rescanProgressReport.TotalHeadersToScan)
+
+			rescanProgressReport.RescanProgress = int32(math.Round(rescanRate * 100))
+			estimatedTotalRescanTime := int64(math.Round(float64(elapsedRescanTime) / rescanRate))
+			rescanProgressReport.RescanTimeRemaining = estimatedTotalRescanTime - elapsedRescanTime
+
+			rescanProgressReport.GeneralSyncProgress = &GeneralSyncProgress{
+				TotalSyncProgress:         rescanProgressReport.RescanProgress,
+				TotalTimeRemainingSeconds: rescanProgressReport.RescanTimeRemaining,
+			}
```
itswisdomagain: This data is basically being repeated. Is this `rescanProgressReport.GeneralSyncProgress` necessary?

rescan.go
```diff
@@ -0,0 +1,116 @@
...
+	if wallet == nil {
+		return errors.E(ErrNotExist)
+	}
+
+	netBackend, err := wallet.internal.NetworkBackend()
+	if err != nil {
+		return errors.E(ErrNotConnected)
+	}
+
+	if mw.IsRescanning() || !mw.IsSynced() {
```
itswisdomagain: `mw.IsRescanning()` prevents being able to perform simultaneous rescans on multiple wallets. Is this intended? put another way, can't we do a `wallet.RescanBlocks()` method instead of a `mw.RescanBlocks(walletID)` method?

rescan.go
```diff
@@ -0,0 +1,116 @@
...
+			mw.syncData.rescanning = false
+			mw.syncData.mu.Unlock()
+		}()
+
+		mw.syncData.mu.Lock()
+		mw.syncData.rescanning = true
+		mw.syncData.mu.Unlock()
+
+		if mw.blocksRescanProgressListener != nil {
+			mw.blocksRescanProgressListener.OnBlocksRescanStarted()
```
itswisdomagain: This does not broadcast the ID for the wallet currently being rescanned. On an app with multiple wallets, how does the app know the wallet for which this `OnBlocksRescanStarted` notification is triggered? Same applies to the `OnBlocksRescanEnded` notification.

rescan.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		ctx, cancel := wallet.shutdownContextWithCancel()
+		mw.syncData.cancelRescan = cancel
```
itswisdomagain: Needs a mutex lock, I believe.

rescan.go
```diff
@@ -0,0 +1,116 @@
...
+				TotalTimeRemainingSeconds: rescanProgressReport.RescanTimeRemaining,
+			}
+
+			if mw.blocksRescanProgressListener != nil {
+				mw.blocksRescanProgressListener.OnBlocksRescanProgress(rescanProgressReport)
+			}
+
+			select {
+			case <-ctx.Done():
+				log.Info("Rescan canceled through context")
```
itswisdomagain: Wouldn't we need to trigger the `OnBlocksRescanEnded` notification with `ctx.Error()`?

rescan.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}()
+
+	return nil
+}
+
+func (mw *MultiWallet) CancelRescan() {
+	if mw.syncData.cancelRescan != nil {
+		mw.syncData.cancelRescan()
+		mw.syncData.cancelRescan = nil
```
itswisdomagain: mutex lock?

rescan.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if mw.IsRescanning() || !mw.IsSynced() {
+		return errors.E(ErrInvalid)
+	}
+
+	go func() {
+		defer func() {
+			mw.syncData.mu.Lock()
+			mw.syncData.rescanning = false
```
itswisdomagain: Should probably also set `mw.syncData.cancelRescan = nil`

txindex/save.go
```diff
@@ -40,3 +40,12 @@ func (db *DB) SaveLastIndexPoint(endBlockHeight int32) error {
 	}
 	return nil
 }
+
+func (db *DB) ClearSavedData(emptyTxPointer interface{}) error {
```
itswisdomagain: `ClearTransactionsIndex` or `ClearSavedTransactions` instead. Clearing saved data should not ordinarily save last index point as 0, but this method does so. It makes more sense if we call this method `ClearSavedTransactions` as it explains why we also set the last index point to 0 after dropping the db.

rescan.go
C-ollins: The app doesn’t perform simultaneously rescan.

rescan.go
C-ollins: Yes, kotlin requires it to be non-nil.

rescan.go
```diff
@@ -0,0 +1,122 @@
...
+			mw.syncData.mu.Unlock()
+		}()
+
+		mw.syncData.mu.Lock()
+
+		mw.syncData.rescanning = true
+
+		if mw.blocksRescanProgressListener != nil {
+			mw.blocksRescanProgressListener.OnBlocksRescanStarted(walletID)
+		}
```
itswisdomagain: Let's move this to after unlocking the mutex to prevent potentially holding the lock for a while. I believe since the listener is not being invoked in a background thread, this would block until the listener completes responding to the notification.

rescan.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+
+			if mw.blocksRescanProgressListener != nil {
+				mw.blocksRescanProgressListener.OnBlocksRescanProgress(rescanProgressReport)
+			}
+
+			select {
+			case <-ctx.Done():
+				log.Info("Rescan canceled through context")
+				mw.blocksRescanProgressListener.OnBlocksRescanEnded(walletID, ctx.Err())
```
itswisdomagain: Missing listener nil check.

rescan.go
```diff
@@ -0,0 +1,128 @@
...
+
+			select {
+			case <-ctx.Done():
+				log.Info("Rescan canceled through context")
+
+				if ctx.Err() != nil && ctx.Err() != context.Canceled {
+					mw.blocksRescanProgressListener.OnBlocksRescanEnded(walletID, ctx.Err())
+				} else {
+					mw.blocksRescanProgressListener.OnBlocksRescanEnded(walletID, nil)
+				}
```
itswisdomagain: This should all be wrapped in a `if mw.blocksRescanProgressListener != nil` block.

PR #80, opened 2020-01-02
spv: use rescan point from lowest chain wallet

C-ollins:
This will avoid a crash caused when the lowest chain wallet doesn't have the block returned from the lowest rescan point.

### DIFF
--- a/spv/sync.go
+++ b/spv/sync.go
@@ -750,6 +750,11 @@ func (s *Syncer) lowestRescanPoint(ctx context.Context) (*chainhash.Hash, error)
 		}
 
 		if rescanPoint == nil {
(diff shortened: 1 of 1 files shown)


spv/sync.go
```diff
@@ -750,6 +750,11 @@ func (s *Syncer) lowestRescanPoint(ctx context.Context) (*chainhash.Hash, error)
 		}
 
 		if rescanPoint == nil {
+			hash, tip := w.MainChainTip(ctx)
+			if tip < rescanBlockHeight {
```
itswisdomagain:
```suggestion
			if tip < rescanBlockHeight || rescanBlockHeight == -1  {
```

Issue #91, opened 2020-01-17
NewMultiWallet panics when given an empty string as netType


RogueElement:
It also fails when given an empty string as rootDir.

itswisdomagain:
Shouldn't panic when given an empty string as netType. Compare #90. Should probably inspect further if calling `NewMultiwallet` with empty string netType causes panic on a client app.

PR #92, opened 2020-01-18
multi: fix dcrwallet/errors usage

### LINKED ISSUE #90: NewMultiWallet does not return a readable error
How to reproduce? Give `NewMultiWallet` invalid arguments and observe what it returns.

Why? Because the code wrongly uses `errors.E`. [This](https://github.com/raedahgroup/dcrlibwallet/blob/master/multiwallet.go#L46) returns just `netType`.

C-ollins:
Errors were wrongly formatted and it returned an unreadable error. The fixes it by using the correct error formatting function.
Closes #90 & #91


log.go
```diff
@@ -89,30 +89,29 @@ var subsystemLoggers = map[string]slog.Logger{
...
 // create roll files in the same directory.  It must be called before the
 // package-global log rotater variables are used.
-func initLogRotator(logFile string) {
+func initLogRotator(logFile string) error {
 	logDir, _ := filepath.Split(logFile)
 	err := os.MkdirAll(logDir, 0700)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to create log directory: %v\n", err)
-		os.Exit(1)
+		return errors.Errorf("failed to create log directory: %v", err)
```
RogueElement: Why not use `fmt.Errorf`?

log.go
```diff
@@ -89,30 +89,29 @@ var subsystemLoggers = map[string]slog.Logger{
...
 
 	logRotator = r
+	return nil
 }
 
 // RegisterLogger should be called before logRotator is initialized.
 func RegisterLogger(tag string) (slog.Logger, error) {
 	if logRotator != nil {
-		return nil, fmt.Errorf("cannot register logger after log rotator is initialized")
+		return nil, errors.E("cannot register logger after log rotator is initialized")
```
RogueElement: Again, why not use `fmt`?

multiwallet.go
```diff
@@ -37,13 +37,25 @@ type MultiWallet struct {
...
+	}
+
+	if netType == "" {
+		netType = chaincfg.MainNetParams().Name
+	}
+
 	chainParams := utils.ChainParams(netType)
 	if chainParams == nil {
-		return nil, errors.E("unsupported network type: %s", netType)
+		return nil, errors.Errorf("unsupported network type: %s", netType)
```
RogueElement: what is wrong with `fmt.Errorf`?

multiwallet.go
```diff
@@ -37,13 +37,25 @@ type MultiWallet struct {
...
 func NewMultiWallet(rootDir, dbDriver, netType string) (*MultiWallet, error) {
-	rootDir = filepath.Join(rootDir, netType)
-	initLogRotator(filepath.Join(rootDir, logFileName))
 	errors.Separator = ":: "
 
+	if rootDir == "" {
+		return nil, errors.E(errors.Invalid, "rootDir must be a valid folder path")
+	}
+
+	if netType == "" {
```
RogueElement: what if the person just forgot to put a proper string? why use Mainnet?

multiwallet.go
```diff
@@ -37,13 +37,25 @@ type MultiWallet struct {
 }
 
 func NewMultiWallet(rootDir, dbDriver, netType string) (*MultiWallet, error) {
-	rootDir = filepath.Join(rootDir, netType)
-	initLogRotator(filepath.Join(rootDir, logFileName))
 	errors.Separator = ":: "
 
+	if rootDir == "" {
```
RogueElement: what if `rootDir` is an not `""`, but an invalid path string like `"/|\"`?

syncnotification.go
```diff
@@ -8,12 +8,6 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-const (
-	SyncStateStart    = "start"
-	SyncStateProgress = "progress"
-	SyncStateFinish   = "finish"
-)
-
```
RogueElement: Was this not used?

C-ollins: Yes

multiwallet.go
C-ollins: github.com/decred/dcrwallet/errors is used all through the project

log.go
C-ollins: github.com/decred/dcrwallet/errors is used all through the project

log.go
C-ollins: github.com/decred/dcrwallet/errors is used all through the project

log.go
RogueElement: alright

RogueElement: But can't we use concrete error types? Searching the error strings for the exact error isn't that nice.

log.go
```diff
@@ -89,30 +88,24 @@ var subsystemLoggers = map[string]slog.Logger{
...
 // RegisterLogger should be called before logRotator is initialized.
 func RegisterLogger(tag string) (slog.Logger, error) {
 	if logRotator != nil {
-		return nil, fmt.Errorf("cannot register logger after log rotator is initialized")
+		return nil, errors.E(ErrLogRotatorAlreadyInitialized)
 	}
 
 	if _, exists := subsystemLoggers[tag]; exists {
-		return nil, fmt.Errorf("logger already registered for tag: %s", tag)
+		return nil, errors.E(ErrLoggerAlreadyRegistered)
```
itswisdomagain: Why do these need defined string constants? Do we anticipate the frontend checking for these particular error cases and displaying the appropriate error message to the user? Wouldn't it be simpler if the error message returned from this and similar functions are more human-readable?

utils/netparams.go
```diff
@@ -4,20 +4,21 @@ import (
...
 	switch strings.ToLower(netType) {
 	case strings.ToLower(mainnetParams.Name):
-		return mainnetParams
+		return mainnetParams, nil
 	case strings.ToLower(testnetParams.Name):
-		return testnetParams
+		return testnetParams, nil
 	default:
-		return nil
+		return nil, errors.New("net type must be either mainnet or testnet3")
```
itswisdomagain: I'd just say "invalid net type" as we can extend this to support other net types in the future. I remember it was discussed once to support simnet (particularly for godcr, I think).

multiwallet.go
```diff
@@ -37,23 +37,32 @@ type MultiWallet struct {
...
+		return nil, errors.Errorf("failed to init logRotator: %v", err.Error())
 	}
 
 	walletsDb, err := storm.Open(filepath.Join(rootDir, walletsDbName))
 	if err != nil {
 		log.Errorf("Error opening wallets database: %s", err.Error())
 		if err == bolt.ErrTimeout {
 			// timeout error occurs if storm fails to acquire a lock on the database file
-			return nil, errors.E("wallets database is in use by another process")
+			return nil, errors.E(ErrWalletDatabaseInUse)
```
itswisdomagain: +1 for this. Client apps can now check the error returned against this constant and easily recognize and handle issues where a wallet db is already opened. As mentioned earlier, don't think similar constant stuff is needed for the log rotator errors.

multiwallet_test.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// TestNewMultiWallet tests if NewMultiWallet will return an error for valid
+// paths or not return an error for invalid paths
+func TestNewMultiWallet(t *testing.T) {
+	f := func(rootDir string) bool {
+		canUse := canUseDir(rootDir)
+		_, err := NewMultiWallet(rootDir, "", "testnet")
+		if canUse {
+			return err == nil
```
itswisdomagain: It is possible that the directory is usable (i.e. it exists or has been created if it doesn't exist) but some other error is returned. In such a case, this would return false and that would not be a representation that the directory is not usable.

Summary: do not expect error to be nil because the directory is usable, but expect err to NOT be nil if directory is not usable.

On another note, do we really even need this test? I believe test cases should be created from user stories and some defined acceptance criteria, not nit-picky.

multiwallet_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	. "github.com/raedahgroup/dcrlibwallet"
+)
+
+// canUseDir returns true if the program can create the directory
+// if it doesn't exist or the directory is empty if it exists.
+// If will delete the directory if it is empty.
+func canUseDir(directory string) bool {
+	if os.MkdirAll(directory, os.ModePerm) == nil {
+		return os.Remove(directory) == nil
```
itswisdomagain: Considering that this function is used to test the `NewMultiwallet` function, this should return nil. `NewMultiWallet` only attempts to create the directory if it does not exist, it does not remove any existing content from the directory - which is important since the root dir may contain a `wallet.db` file that will/should be linked to the new multiwallet instance, AFTER, the instance is created.

multiwallet.go
```diff
@@ -37,23 +37,32 @@ type MultiWallet struct {
...
-	chainParams := utils.ChainParams(netType)
-	if chainParams == nil {
-		return nil, errors.E("unsupported network type: %s", netType)
+	chainParams, err := utils.ChainParams(netType)
+	if err != nil {
+		return nil, err
+	}
+
+	rootDir = filepath.Join(rootDir, netType)
+	err = os.MkdirAll(rootDir, 0700)
```
itswisdomagain:
```suggestion
	err = os.MkdirAll(rootDir, os.ModePerm)
```

multiwallet_test.go
RogueElement: It's Quality Assurance, if it can be done, it will be done and it should behave in a expected manner.
What other error can there be?

multiwallet_test.go
RogueElement: Yeah, I sure knew that since it was documented somewhere...

multiwallet.go
RogueElement: No no no. Why are you using strings? What exactly is the point of returning an error type if all branches return a string? Either use an error type or use a string.

multiwallet_test.go
RogueElement: And the thing is that this removes the directory if it exists but is empty (like the code comment said) so all values for which `canUseDir` returns `true` are valid paths that do not exist.

multiwallet.go
C-ollins: The error contains an exported string variable so it can be referenced from mobile to determine what error was thrown. It's returning an error type to trigger the exception in a try-catch block(for mobile) and gomobile would ignore a function with 2 return types.

> Any function type all of whose parameters and results have
>   supported types. Functions must return either no results,
>   one result, or two results where the type of the second is
>   the built-in 'error' type.

https://godoc.org/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions.

log.go
C-ollins: We're considering a feature where we report all errors in the frontend, declaring error constants would enable the frontend determine what error was thrown.

itswisdomagain: Apologies. I had mixed this up. These error constants would be valuable to the frontend.

multiwallet.go
itswisdomagain: Apologies for the confusion, I meant to return an error with a string literal rather than a string constant in the log rotator function. Of course, it is atypical to return a string instead of an error.

In any case, I've also clarified my comments on the log rotator error constants.

multiwallet_test.go
itswisdomagain: Ah I see. I guess that's the point I'm trying to make. A path may also be valid if it exists and is not empty. This function would return an error for any valid path that is not empty.

multiwallet_test.go
itswisdomagain: A simple scan of the `NewMultiwallet` function shows multiple sites where an error could be returned, theoretically speaking. Again, it is possible that the directory is usable (i.e. it exists or has been created if it doesn't exist) but some other error is returned.

itswisdomagain: For the bit about QA, I respectfully disagree. The statement, "if it can be done, it will be done" does not and should not apply to writing tests. 
Doing so would easily amount to a waste of programmer time.

Imagine writing test to ensure that a function properly returns the sum of 2 numbers. Err, well maybe we should write such a test but like I said above, you don't test basically every little bit of code to ensure it works as expected. You write test cases following user stories and acceptance criteria.

Instead of writing a test to check `if NewMultiWallet will return an error for valid paths or not return an error for invalid paths`, we probably should be testing more relevant/sensitive methods and functions. Imo, a good test would be - `NewMultiWallet` should return an error for invalid arguments and no error for valid arguments. Testing specifically for path validity is nit-picky in my opinion.

That said, I have no reason to not merge this if the test code is corrected as requested in the initial review. I just think we can do better.

RogueElement: Where are these user stories and acceptance criteria you speak of?

RogueElement: Path validity is input validity. It should return an error for invalid arguments(cannot use the path argument) and no error for valid arguments (can use the path argument). That's exactly what the test checks...

multiwallet_test.go
RogueElement: It doesn't return an error...

multiwallet_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	. "github.com/raedahgroup/dcrlibwallet"
+)
+
+// canUseDir returns true if the program can create the directory
+// if it doesn't exist or the directory is empty if it exists.
+// If will delete the directory if it is empty.
+func canUseDir(directory string) bool {
+	if os.MkdirAll(directory, os.ModePerm) == nil {
+		return os.Remove(directory) != nil
```
itswisdomagain: Still won't work. `os.Remove` will be nil if the directory is empty. Doesn't make the directory unusable. My recommendation:
```suggestion
		return true
```

RogueElement: It passes...

RogueElement: Why wouldn't an empty dir be usable?

itswisdomagain: It is usable. But following your code, an empty directory will return false here because `os.Remove` will be nil.

RogueElement: It can't be empty if it got here, `os.MkdirAll` creates it?

itswisdomagain: `os.MkdirAll` creates an empty directory.

RogueElement: Oh yeah, was confused for a sec

.github/workflows/build.yml
```diff
@@ -1,36 +1,23 @@
...
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: go mod download
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
 
-      - name: Lint
+      - name: Build, run tests and vet
         run: |
-          export PATH=$PATH:~/go/bin
-          ./run_tests.sh lint
```
itswisdomagain: With this change, we'd need to remove the `./run_tests.sh` file as well. However, I'd recommend keeping it. Should probably rename it though, to say `./lint.sh`.

.github/workflows/build.yml
```diff
@@ -1,36 +1,23 @@
...
-        uses: actions/cache@v1
-        id: cache
-        with:
-          path: ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-
-
-      - name: Install linter
-        run: "curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.22.2"
```
itswisdomagain: If we're keeping the linter sh file, we'll need to retain this as well.

.github/workflows/build.yml
```diff
@@ -1,36 +1,23 @@
...
-      - name: Lint
+      - name: Build, run tests and vet
         run: |
-          export PATH=$PATH:~/go/bin
-          ./run_tests.sh lint
+          go build
+          go test
+          go vet
+        env:
+          GO111MODULE: "on"
```
itswisdomagain: While we're here, I'd recommend moving this to before `run`.

multiwallet_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"os"
+	"testing"
+	"testing/quick"
+
+	. "github.com/raedahgroup/dcrlibwallet"
+)
+
+// canUseDir returns true if the program can create the directory
+// It will create the directory if it can
```
itswisdomagain: This comment is unnecessary. The previous line covers it.

multiwallet_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"os"
+	"testing"
+	"testing/quick"
+
+	. "github.com/raedahgroup/dcrlibwallet"
+)
+
+// canUseDir returns true if the program can create the directory
```
itswisdomagain:
```suggestion
// canUseDir returns true if the program can create the directory.
```

multiwallet_test.go
```diff
@@ -0,0 +1,31 @@
...
+// It will create the directory if it can
+func canUseDir(directory string) bool {
+	return os.MkdirAll(directory, os.ModePerm) == nil
+}
+
+// TestNewMultiWalletPath checks that NewMultiWallet returns an error for invalid
+// paths or no error for valid paths
```
itswisdomagain:
```suggestion
// paths
```
Valid paths may still return an error.

multiwallet_test.go
```diff
@@ -0,0 +1,31 @@
...
+func TestNewMultiWallet(t *testing.T) {
+	f := func(rootDir string) bool {
+		canUse := canUseDir(rootDir)
+		_, err := NewMultiWallet(rootDir, "", "testnet")
+		if canUse {
+			return err == nil
```
itswisdomagain: As has been explained previously, this test is invalid because `canUse` can be true, yet `err != nil`. Using a valid path DOES NOT always mean that `NewMultiWallet` will return no errors; it might, it might not.

It all boils down to what is being tested in this function. I was even gonna recommend renaming the test function as this testing code does not perform a complete test of the `NewMultiWallet` function. As indicated in the comment, it all tests path validity.

If during a test, the path is valid AND an error is returned, this test will fail when it should not. Reason: if a test fails, it means there's a problem somewhere - either with the code being tested or the testing code itself. In the scenario highlighted above, the test would fail, not necessarily because there is a problem with the `NewMultiWallet` function, but because there's a problem with this testing code.

multiwallet_test.go
```diff
@@ -0,0 +1,31 @@
...
+		}
+		return err != nil
+	}
```
itswisdomagain: As have been said previously, this test is buggy. I made some changes to the `f` function, ran the test a couple times to get 2 set of results (one passed, one failed) which reveal some issues with this testing code. The log outputs displayed for both results are truncated.

#### Modified the `f` function to print somewhat detailed reports
```
f := func(rootDir string) bool {
	canUse := canUseDir(rootDir)
	_, err := NewMultiWallet(rootDir, "", "testnet")

	if canUse {
		passed := err == nil

		var report string
		if passed {
			report = fmt.Sprintf("GOOD: %U is a valid path and we got no errors. Yay!", rootDir)
		} else {
			report = fmt.Sprintf("BAD: %U is a valid path but we got an error: %s.", rootDir, err.Error())
		}
		println(report)

		return passed
	}

	passed := err != nil

	var report string
	if passed {
		report = fmt.Sprintf("GOOD: %U is NOT a valid path, so we got an error: %s.", rootDir, err.Error())
	} else {
		report = fmt.Sprintf("BAD: %U is NOT a valid path but we got no errors. Weird!", rootDir)
	}
	println(report)

	return passed
}
```

#### First test result + log output (TEST PASSED with exit code 0, but should it have?)
```
GOOD: "\U000cbb31\U0006ecc1⋯\U000859ac𮗚\U0008311f\U000df8a3\U00076647" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U0008ddd5𡢙\U000c97e4\U00106c29\U000ac4b7̌\U000f4f23\U0004faa2\U000547ee\U0003e399\U00046378\U0001f269\U0010a688𬡚\U0009449d𗢷" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U0001ecac\U00095bba奕\U000782e0\U000ceb5b\U000b03f1\U0007d731\U00092766𨞈\U000a9ae0\U000aebf2" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U00066d7c\U000c02a8" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U0005214c\U00075f48\U0004562f\U0004ee05" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U00054699\U00097366\U0004709a" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U00085c61\U00042df4\U000149ee\U0006f80c\U000ada6e\U00067402" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U00050e5a\U00043f9f贉\U0007b60f\U0005f314\U000bd61d\U0005847b�\U000622ce\U000671f0췢\U0010a726\U000e4da9" is NOT a valid path, so we got an error: invalid net type.

Process finished with exit code 0
```
Analysis
- All test parameters passed to `f` were invalid paths. So it is expected that `NewMultiWallet` will return an error; and it did! But the error returned was `invalid net type` instead of a `PathError`. Agreed, if the netType was indeed valid, the error would have been a `PathError`, but this just goes to emphasize that the `NewMultiWallet` function can return an error whether or not a valid path is provided.
- If a valid netType was passed, the error would look like `GOOD: "\U0007af92\U0010ac8d\U0009d6ae" is NOT a valid path, so we got an error: failed to create rootDir: mkdir 񺾒􊲍򝚮: illegal byte sequence.` This kind of error should be what is used in determining whether the test should pass or fail. Any error shouldn't do.

#### Second test result + log output (TEST FAILED with exit code 1, but should it have?)
```
GOOD: "" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U000cd531\U00030f92\U0006d2cf\U000f2b31" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U001018f1\U000c57d4𨃁\U00101e3f\U00107546" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U000d9d16" is NOT a valid path, so we got an error: invalid net type.
GOOD: "" is NOT a valid path, so we got an error: invalid net type.
GOOD: "" is NOT a valid path, so we got an error: invalid net type.
GOOD: "𗰶𨺈䵄\U000eca4a\U000c2b42" is NOT a valid path, so we got an error: invalid net type.
GOOD: "\U0006d593𤣯\U000f1629" is NOT a valid path, so we got an error: invalid net type.
BAD: "\ue91b" is a valid path but we got an error: invalid net type.
    multiwallet_test.go:52: #47: failed on input "\ue91b"

Process finished with exit code 1
```
Analysis
- The test failed even though the path was valid. This shouldn't be.
- An error may be returned even when the path is valid. Therefore, it is incorrect to expect no errors simply because the path is valid.
- This the reason the last test run on this PR failed. It shouldn't have.

#### Other observations:
- It turns out that relative paths are supported but whether or not this is desired is yet to be seen.
- `canUseDir` creates the directory if it does not exist, should also ensure that the directory is deleted IF it was created. Should not delete the directory if it was already existing.

#### An interesting edge case
If the test code is corrected to use a valid `netType` (say `testnet3`), the test may still fail with the following error:
```
BAD: "" is NOT a valid path but we got no errors. Weird!
2020-01-22 12:20:57.101 [INF] DLWL: Loaded 0 wallets
    multiwallet_test.go:52: #90: failed on input ""

Process finished with exit code 1
```
This means that the `NewMultiWallet` function had no issues using the directory (no errors returned from the function) but `canUseDir` said the directory is not valid. Who's wrong?

PR #94, opened 2020-01-22
Refactor syncData mutex to prevent potentially long locks

itswisdomagain:
Ensures that access to ongoing sync data uses mutex locks held for as little time as possible to prevent delays when some other operation attempts to access ongoing sync data.

It is particularly important to unlock the syncData mutex before triggering sync progress listeners registered by client apps, since some of such listeners may perform long-running operations or even try to access ongoing sync data - in which case, an attempt will be made to lock the syncData mutex. In the latter scenario, the attempt to lock the mutex would lead to a deadlock if the mutex was not previously unlocked before the client-registered listener was invoked.

Also fixed a bug with `cancelSync` where initial syncing may have completed and the method is called to disconnect the wallets from the network backend.
https://github.com/raedahgroup/dcrlibwallet/blob/064d4c3c9998da7c04a9841726b16c34ced7e4fa/sync.go#L252-L258
The problem observed is that `mw.syncData.cancelSync` is not nil (even after initial sync completes because the wallets remain connected) but trying to access `mw.syncData.addressDiscoveryCompletedOrCanceled` panics because `mw.syncData.activeSyncData` would be nil (after initial sync completes) and `mw.syncData.addressDiscoveryCompletedOrCanceled` is actually `mw.syncData.activeSyncData.addressDiscoveryCompletedOrCanceled`.

### DIFF
--- a/sync.go
+++ b/sync.go
@@ -37,18 +37,15 @@ type syncData struct {
 
 // reading/writing of properties of this struct are protected by syncData.mu.
 type activeSyncData struct {
-	targetTimePerBlock int32
-
 	syncStage int32
 
 	headersFetchProgress     HeadersFetchProgressReport
 	addressDiscoveryProgress AddressDiscoveryProgressReport
 	headersRescanProgress    HeadersRescanProgressReport
 
-	beginFetchTimeStamp      int64
-	totalFetchedHeadersCount int32
-	startHeaderHeight        int32
-	headersFetchTimeSpent    int64
+	beginFetchTimeStamp   int64
+	startHeaderHeight     int32
+	headersFetchTimeSpent int64
 
 	addressDiscoveryStartTime int64
 	totalDiscoveryTimeSpent   int64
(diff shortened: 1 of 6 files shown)


sync.go
```diff
@@ -249,17 +262,18 @@ func (mw *MultiWallet) RestartSpvSync() error {
 }
 
 func (mw *MultiWallet) CancelSync() {
-	if mw.syncData.cancelSync != nil {
+	if mw.IsSyncing() {
```
itswisdomagain: This is not quite correct. Syncing may have completed but this method is called to disconnect the wallet from the network backend. The actual problem here is that `mw.syncData.cancelSync` may not be nil (even after initial sync completes) but `mw.syncData.addressDiscoveryCompletedOrCanceled` would be accessible because `mw.syncData.activeSyncData` would be nil and `mw.syncData.addressDiscoveryCompletedOrCanceled` is actually `mw.syncData.activeSyncData.addressDiscoveryCompletedOrCanceled`.

PR #97, opened 2020-01-24
Upgrade v1.1.0 modules to Decred v1.5.0 release

C-ollins:
All of dcrd & dcrwallet modules were bumped to the module versions used in Decred v1.5.0 release, and the library changes were also implemented in affected files. The modules in the current release are outdated.


addresshelper/helper.go
```diff
@@ -2,13 +2,15 @@ package addresshelper
...
-	"github.com/decred/dcrd/dcrutil"
-	"github.com/decred/dcrd/txscript"
+	"github.com/decred/dcrd/chaincfg/v2"
+	"github.com/decred/dcrd/dcrutil/v2"
+	"github.com/decred/dcrd/txscript/v2"
 )
 
-func PkScript(address string) ([]byte, error) {
-	addr, err := dcrutil.DecodeAddress(address)
+const ScriptVersion = 0
```
itswisdomagain:
```suggestion
const DefaultScriptVersion = 0
```

sync.go
```diff
@@ -201,13 +195,12 @@ func (lw *LibWallet) SpvSync(peerAddresses string) error {
...
 	if len(validPeerAddresses) > 0 {
-		syncer.SetPersistantPeers(validPeerAddresses)
+		syncer.SetPersistentPeers(validPeerAddresses)
 	}
 
 	loadedWallet.SetNetworkBackend(syncer)
-	lw.walletLoader.SetNetworkBackend(syncer)
 
-	ctx, cancel := lw.contextWithShutdownCancel(context.Background())
+	ctx, cancel := context.WithCancel(context.Background())
```
itswisdomagain: With this approach, sync won't get canceled when an interrupt signal is received.
To fix, also execute `lw.CancelSync()` (if `lw.CancelSync != nil` of course) after canceling all contexts created with `lw.shutdownContext()`. This should be done when the `lw.shuttingDown` signal is received in the go routine in `lw.listenForShutdown()`.

This means that there would be no need to do `lw.CancelSync()` in `lw.Shutdown()` because `lw.Shutdown()` sends a signal to `lw.shuttingDown` to cancel all contexts created with `lw.shutdownContext()`.

sync.go
```diff
@@ -431,6 +381,11 @@ func (lw *LibWallet) RescanBlocks() error {
 			}
 		}
 
+		err = lw.reindexTransactions()
+		if err != nil {
+			log.Errorf("Error clearing saved transactions: %v", err)
```
itswisdomagain:
```suggestion
			log.Errorf("Error re-indexing transactions: %v", err)
```

txhelper/outputs.go
```diff
@@ -33,7 +32,7 @@ func MakeTxOutput(destination TransactionDestination) (output *wire.TxOut, err e
 
 	output = &wire.TxOut{
 		Value:    int64(amountInAtom),
-		Version:  txscript.DefaultScriptVersion,
+		Version:  addresshelper.ScriptVersion,
```
itswisdomagain: I think the `DefaultScriptVersion` should be defined in this package instead of the `addresshelper` package.

dcrlibwallet.go
```diff
@@ -115,10 +116,6 @@ func (lw *LibWallet) Shutdown() {
 	// Trigger shuttingDown signal to cancel all contexts created with `contextWithShutdownCancel`.
```
itswisdomagain:
```suggestion
	// Trigger shuttingDown signal to cancel all contexts created with `lw.shutdownContext()`.
```

txhelper/outputs.go
C-ollins: That would cause an import cycle error.

dcrlibwallet.go
```diff
@@ -112,13 +113,9 @@ func newLibWallet(walletDataDir, walletDbDriver string, activeNet *netparams.Par
...
 
-	// Trigger shuttingDown signal to cancel all contexts created with `contextWithShutdownCancel`.
+	// Trigger shuttingDown signal to cancel all contexts created with `lw.shutdownContext()`.
 	lw.shuttingDown <- true
 
-	if lw.rpcClient != nil {
-		lw.rpcClient.Stop()
-	}
-
 	lw.CancelSync()
```
itswisdomagain: Should at least wrap this in `if lw.CancelSync != nil`. But as recommended in a previous review, this nil check and `cancelSync()` execution should be done in the go routine that listens for `lw.shuttingDown` signals.

The idea is to cancel all created shutdown contexts in that go routine, and also cancel sync if cancelSync != nil.

PR #112, opened 2020-02-28
Account test

ademuanthony:
This is PR add tests for account operations in the accounts.go file.

An interface `walletint.InternalWallet` is extracted from the tightly coupled `dcrwallet/wallet.Wallet` to isolate the dependency. The `InternalWallet` is used as the type for the `internal` field of the `Wallet` struct and it is injected as an optional parameter to the `NewMultiWallet` function. 
The test uses a fake implementation of the interface as its internal wallet to aid deterministic assertion. 
A wrapper is created around the `dcrwallet/wallet.Wallet` struct. This wrapper is made to satisfy the `walletint.InternalWallet` interface and  is used as the internal wallet if the user did not pass the `walletint.InternalWallet` optional parameter to `NewMultiWallet`.

The PR use github.com/onsi/ginkgo and github.com/onsi/gomega as introduced by @RogueElement in his on going PR #96 
github.com/golang/mock is used as mocking framework

image


account_test.go
```diff
@@ -0,0 +1,218 @@
...
+				By("returning nonempty accounts string")
+				Expect(res).ToNot(BeEquivalentTo(""))
+			})
+		})
+	})
+
+	Describe("GetAccount(accountNumber int32, requiredConfirmations int32) (*Account, error)", func() {
+		Context("when called with the right parameters", func() {
+			It("should get account", func() {
+
```
RogueElement: I'm not sure why this extra whitespace is here

account_test.go
```diff
@@ -0,0 +1,218 @@
...
+	Describe("GetAccountBalance(accountNumber int32, requiredConfirmations int32) (*Balance, error)", func() {
+		Context("when called", func() {
+			It("should get the account balance", func() {
+
+				balance, err := wallet.GetAccountBalance(0, 0)
+				By("returning a nil error")
+				Expect(err).To(BeNil())
+				By("returning the expected balance")
+				Expect(balance).ToNot(BeNil())
```
RogueElement: Just not `nil` is not the expected balance. If the wallet was just created, then the balance should be equivalent to 0.

account_test.go
```diff
@@ -0,0 +1,218 @@
...
+	var (
+		wallet *Wallet
+	)
+
+	BeforeEach(func() {
+		multi, err := NewMultiWallet(rootDir, "", "testnet3")
+		Expect(err).To(BeNil())
+		Expect(multi).ToNot(BeNil())
+
+		wallet, err = multi.CreateNewWallet("0000", 0)
```
RogueElement: Randomizing the password and passtype seems like a good idea. (it's not worth the stress if it takes anymore dev/running time)

account_test.go
```diff
@@ -0,0 +1,290 @@
...
+		Expect(err).To(BeNil())
+		Expect(wallet).ToNot(BeNil())
+		internalWallet = wallet.InternalWallet()
+	})
+
+	AfterSuite(func() {
+		if wallet != nil {
+			wallet.Shutdown()
+		}
+	})
```
itswisdomagain, line 48: Since `BeforeSuite` and `AfterSuite` run before and after ALL test blocks, I think it's more readable to place these outside the `Describe` block - they should have no parent block.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,34 @@
+name: Tests
```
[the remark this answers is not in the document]
ademuanthony: There was a problem with running the test on windows. But that seems okay now. This can be removed, I guess

Issue #110, opened 2020-02-26
Prevent gap exhaustion in change generation


raedah:
There was a report of this occurring on mobile. After the issue is confirmed on mobile, it needs to be resolved.

https://github.com/decred/dcrwallet/issues/1622
https://github.com/decred/decrediton/pull/2358

C-ollins:
Closed via #115

PR #113, opened 2020-03-01
send: prevent address gap limit exhaustion in change generation

C-ollins:
This fixes a gap limit exhaustion error caused by calling `ConstructTransaction` multiple times which causes the wallet to keep generating change addresses until exhaustion. This error is fixed by using gap policy 'wrap' which configures the address generation method to wrap around to a previously returned address instead of erroring or ignoring the gap limit and returning a new unused address. The change source created during `ConstructTransaction` is saved, reused throughout the current session and discarded after publishing transaction.

Addresses #110 

Error Message: 
`[ERR] DLWL: wallet.NewUnsignedTransaction: policy violation:: txauthor.NewUnsignedTransaction: generating next address violates the unused address gap limit policy`

### DIFF
--- a/dcrlibwallet.go
+++ b/dcrlibwallet.go
@@ -32,7 +32,6 @@ import (
 	"github.com/decred/dcrwallet/p2p/v2"
 	"github.com/decred/dcrwallet/spv/v3"
 	"github.com/decred/dcrwallet/wallet/v3"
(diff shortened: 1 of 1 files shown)


itswisdomagain:
Stale due to v1.0.1 release.

PR #115, opened 2020-03-02
send: change NewUnsignedTx to multiwallet method, add TxAuthor helper methods

itswisdomagain:
Change from `wallet.NewUnsignedTx` to `multiwallet.NewUnsignedTx` is done to enable the `NewUnsignedTx` method access the user-configured `SpendUnconfirmedFunds` setting.

Also enable access to the `SpendUnconfirmedFunds` setting from wallet methods and remove `requiredConfirmations` parameters from wallet methods.

### DIFF
--- a/txauthor.go
+++ b/txauthor.go
@@ -16,23 +16,19 @@ import (
 )
 
 type TxAuthor struct {
-	sendFromAccount       uint32
+	sourceWallet          *Wallet
+	sourceAccountNumber   uint32
 	destinations          []TransactionDestination
 	requiredConfirmations int32
-	wallet                *Wallet
 }
 
-func (wallet *Wallet) NewUnsignedTx(sourceAccountNumber, requiredConfirmations int32) *TxAuthor {
+func (mw *MultiWallet) NewUnsignedTx(sourceWallet *Wallet, sourceAccountNumber int32) *TxAuthor {
 	return &TxAuthor{
-		sendFromAccount:       uint32(sourceAccountNumber),
+		sourceWallet:          sourceWallet,
+		sourceAccountNumber:   uint32(sourceAccountNumber),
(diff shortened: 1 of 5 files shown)


txauthor.go
```diff
@@ -16,25 +16,19 @@ import (
...
-	destinations          []TransactionDestination
-	requiredConfirmations int32
-	wallet                *Wallet
+	sourceWallet        *Wallet
+	sourceAccountNumber uint32
+	destinations        []TransactionDestination
 }
 
-func (wallet *Wallet) NewUnsignedTx(sourceAccountNumber, requiredConfirmations int32) *TxAuthor {
+func (mw *MultiWallet) NewUnsignedTx(sourceWallet *Wallet, sourceAccountNumber int32) *TxAuthor {
```
RogueElement: Can we keep `wallet.NewUnsignedTx` and have them both?
And also return nil if the `sourceAccountNumber` is invalid?

itswisdomagain: Keeping `wallet.NewUnsignedTx` would require that the caller pass in the required confirmations. Not desired but possible. I'll look into it.

itswisdomagain: On further thought, @C-ollins has an idea that would work better in making wallet instances access config preferences independently, will wait for such solution before re-trying to make `NewUnsignedTx` a wallet method.

txauthor.go
```diff
@@ -57,6 +53,22 @@ func (tx *TxAuthor) RemoveSendDestination(index int) {
 	}
 }
 
+func (tx *TxAuthor) SendDestination(atIndex int) *TransactionDestination {
+	return &tx.destinations[atIndex]
```
C-ollins: There should be a check here for `atIndex` and this function should also return an error because it'll cause a panic if an invalid index is passed.

PR #118, opened 2020-03-11
log: add exported Log function with tag parameter

C-ollins:
This will help in tracing log source for logs coming from client apps that make use of dcrlibwallet.

### DIFF
--- a/log.go
+++ b/log.go
@@ -142,6 +142,12 @@ func setLogLevel(subsystemID string, logLevel string) {
 	logger.SetLevel(level)
 }
 
+// Log writes a message to the log using LevelInfo
 func Log(m string) {
 	log.Info(m)
+}
+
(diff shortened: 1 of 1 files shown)


log.go
```diff
@@ -142,6 +142,12 @@ func setLogLevel(subsystemID string, logLevel string) {
 	logger.SetLevel(level)
 }
 
+// Log writes a message to the log using LevelInfo
 func Log(m string) {
 	log.Info(m)
 }
+
+// LogT writes a tagged message to the log using LevelInfo
```
itswisdomagain: End both comments with period.
```suggestion
// Log writes a message to the log using LevelInfo.
func Log(m string) {
	log.Info(m)
}

// LogT writes a tagged message to the log using LevelInfo.
```

PR #123, opened 2020-04-08
fix days behind calculation

oshorefueled:
fixes inaccurate result during wallet sync when calculating how far behind in days a wallet is compared to the latest block height.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -214,9 +214,9 @@ func CalculateTotalTimeRemaining(timeRemainingInSeconds int64) string {
 }
 
 func CalculateDaysBehind(lastHeaderTime int64) string {
-	hoursBehind := float64(time.Now().Unix()-lastHeaderTime) / 60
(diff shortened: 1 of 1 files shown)


utils.go
```diff
@@ -214,8 +214,8 @@ func CalculateTotalTimeRemaining(timeRemainingInSeconds int64) string {
 }
 
 func CalculateDaysBehind(lastHeaderTime int64) string {
-	hoursBehind := float64(time.Now().Unix()-lastHeaderTime) / 60
-	daysBehind := int(math.Round(hoursBehind / 24))
+	diff := time.Since(time.Unix(lastHeaderTime, 0))
+	daysBehind := int(math.Round(diff.Hours() / 24))
 	if daysBehind < 1 {
```
RogueElement, line 219: This would be clearer as `daysBehind == 0`.

PR #135, opened 2020-05-10
Fix seed encryption errors

C-ollins:
Addresses an issue where verify seed function returns ErrInvalid for an incorrect passphrase and also fixes a bug where the seed of an already backed up wallet is attempted to be re-encrypted meanwhile the encrypted seed is empty and that throws an uncaught error.

### DIFF
--- a/multiwallet.go
+++ b/multiwallet.go
@@ -626,5 +626,19 @@ func (mw *MultiWallet) ChangePrivatePassphraseForWallet(walletID int, oldPrivate
 
 	wallet.EncryptedSeed = encryptedSeed
 	wallet.PrivatePassphraseType = privatePassphraseType
-	return mw.db.Save(wallet)
+	err = mw.db.Save(wallet)
+	if err != nil {
+		log.Errorf("error saving wallet-[%d] to database after passphrase change: %v", wallet.ID, err)
+
+		err2 := wallet.changePrivatePassphrase(newPrivatePassphrase, oldPrivatePassphrase)
+		if err2 != nil {
+			log.Errorf("error undoing wallet passphrase change: %v", err2)
+			log.Errorf("error wallet passphrase was changed but passphrase type and newly encrypted seed could not be saved: %v", err)
+			return errors.New(ErrSavingWallet)
+		}
+
+		return errors.New(ErrChangingPassphrase)
+	}
+
+	return nil
 }
--- a/multiwallet.go
+++ b/multiwallet.go
@@ -606,13 +606,8 @@ func (mw *MultiWallet) ChangePrivatePassphraseForWallet(walletID int, oldPrivate
 		return errors.New(ErrInvalid)
 	}
 
-	err := wallet.changePrivatePassphrase(oldPrivatePassphrase, newPrivatePassphrase)
-	if err != nil {
-		return translateError(err)
-	}
-
 	if wallet.EncryptedSeed != nil {
-		decryptedSeed, err := decryptWalletSeed(newPrivatePassphrase, wallet.EncryptedSeed)
+		decryptedSeed, err := decryptWalletSeed(oldPrivatePassphrase, wallet.EncryptedSeed)
 		if err != nil {
 			return err
 		}
@@ -623,6 +618,11 @@ func (mw *MultiWallet) ChangePrivatePassphraseForWallet(walletID int, oldPrivate
 		}
 
 		wallet.EncryptedSeed = encrytedSeed
+	}
+
+	err := wallet.changePrivatePassphrase(oldPrivatePassphrase, newPrivatePassphrase)
+	if err != nil {
+		return translateError(err)
 	}
 
 	wallet.PrivatePassphraseType = privatePassphraseType
--- a/multiwallet.go
+++ b/multiwallet.go
@@ -413,6 +413,14 @@ func (mw *MultiWallet) saveNewWallet(wallet *Wallet, setupWallet func() error) (
 		}
 
 		walletDataDir := filepath.Join(mw.rootDir, strconv.Itoa(wallet.ID))
+
+		dirExists, err := fileExists(walletDataDir)
+		if err != nil {
+			return err
+		} else if dirExists {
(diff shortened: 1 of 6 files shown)


multiwallet.go
```diff
@@ -603,17 +611,19 @@ func (mw *MultiWallet) ChangePrivatePassphraseForWallet(walletID int, oldPrivate
 		return translateError(err)
 	}
 
-	decryptedSeed, err := decryptWalletSeed(newPrivatePassphrase, wallet.EncryptedSeed)
-	if err != nil {
-		return err
-	}
+	if wallet.EncryptedSeed != nil {
+		decryptedSeed, err := decryptWalletSeed(newPrivatePassphrase, wallet.EncryptedSeed)
```
itswisdomagain, line 615: Shouldn't the encrypted seed be decrypted with the old passphrase first, then re-encrypted with the new passphrase?

multiwallet_utils.go
```diff
@@ -150,7 +150,8 @@ func decryptWalletSeed(pass []byte, encryptedSeed []byte) (string, error) {
 
 	decryptedSeed, err := secretbox.EasyOpen(encryptedSeed, key)
 	if err != nil {
-		return "", err
+		log.Error(err)
```
itswisdomagain, line 153: Maybe add more info to the logged error to make it searchable and more informative, something like:
```suggestion
		log.Errorf("decryptWalletSeed secretbox.EasyOpen error: %v", err)
```

multiwallet.go
```diff
@@ -413,6 +413,14 @@ func (mw *MultiWallet) saveNewWallet(wallet *Wallet, setupWallet func() error) (
 		}
 
 		walletDataDir := filepath.Join(mw.rootDir, strconv.Itoa(wallet.ID))
+
+		dirExists, err := fileExists(walletDataDir)
+		if err != nil {
+			return err
+		} else if dirExists {
+			os.RemoveAll(walletDataDir)
+		}
```
itswisdomagain, line 422: This is a little disturbing, I'd wager that if a directory exists at this path, then it would hold a wallet database file. Deleting wallet files without user permission / awareness is concerning.

Also, the dir name used for new wallets is a number that's properly incremented by the db to prevent collision, it doesn't even reuse a number even if the wallet that was there is deleted, so we typically should not have this issue, but if we do, I think aborting the operation is the preferable way to go, the user should manually delete the directory that is posing trouble.

It's even possible that the next try from the user won't have that issue because the wallet id would be again incremented.

multiwallet.go
```diff
@@ -598,23 +606,25 @@ func (mw *MultiWallet) ChangePrivatePassphraseForWallet(walletID int, oldPrivate
...
+	err := wallet.changePrivatePassphrase(oldPrivatePassphrase, newPrivatePassphrase)
 	if err != nil {
-		return err
+		return translateError(err)
 	}
 
-	wallet.EncryptedSeed = encrytedSeed
-
 	wallet.PrivatePassphraseType = privatePassphraseType
 	return mw.db.Save(wallet)
```
itswisdomagain, line 629: Should do more than return this error to the caller. Also undo the wallet passphrase change if saving the new encrypted seed to db fails; otherwise the user would not be able to decrypt this wallet's seed subsequently using the new password.

If trying to change back the password errors, we should return an error message that helps the caller understand that there's gonna be some discrepancies between the wallet private passphrase and the seed encryption passphrase.

Something like:
```suggestion
	err := mw.db.Save(wallet)
	if err != nil && wallet.EncryptedSeed != nil {
		log.Errorf("error saving re-encrpyted wallet seed: %v", err)
		err2 := wallet.changePrivatePassphrase(newPrivatePassphrase, oldPrivatePassphrase)
		if err2 != nil {
			log.Errorf("error undoing wallet passphrase change: %v", err2)
			return errors.New("wallet passphrase changed but re-encrypting seed failed")
		}
	}
	return err
```

multiwallet.go
C-ollins: If a panic occurs when creating a wallet due to memory, the wallet folder will remain and the wallet loader won't be able to create the wallet. For mobile, the folders are fully managed by the apps so there should be no scenario of an undocumented folder making it there unless it happened due to panic.

multiwallet.go
C-ollins: Check this out 4f5e49e, It holds the newly encrypted seed in a variable and sets it after changing the wallet private passphrase.

itswisdomagain: It's quite the same thing. In memory, the encrypted seed uses the new password. But if the save fails, the db still retains the seed encrypted with the old password.

multiwallet.go
itswisdomagain: Given that `dcrlibwallet` could be used outside mobile devices, or even in mobile devices with a publicly accessible `rootDir`, the assumption that orphaned wallet dirs are a result of a previous panic becomes a rough assumption at best.

Since `dcrlibwallet` does not ensure that the `rootDir` cannot be modified externally, we should assume that a user could store some valuable information in the conflicting directory and return an error instead of deleting that directory.

utils.go
```diff
@@ -260,6 +260,29 @@ func moveFile(sourcePath, destinationPath string) error {
...
+		err = moveFile(fileName, newName)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	err = backup(fileName)
+	return
```
itswisdomagain, line 283: The nested function doesn't seem to be necessary. Can just be
```
	newName = filename + ".bak"
	exists, err := fileExists(newName)
	if err != nil {
		return err
	} else if exists {
		return backupFile(newName)
	}

	err = moveFile(fileName, newName)
	if err != nil {
		return err
	}

	return nil
```

C-ollins: I added the nested function just so I could return the newName and log it to the user, this implementation ^^^ does not do that.

itswisdomagain: I updated the code snippet.

C-ollins: Updated

PR #138, opened 2020-05-24
prevent more than one wallet from using the same seed

### LINKED ISSUE #132: prevent adding more than 1 wallet with the same seed
It's currently possible to add 2 wallets with the same seed to the app, which makes txs appear double on the recent transactions list .

Should be possible to prevent adding a wallet if it uses the same seed as a previously added wallet when https://github.com/decred/dcrwallet/issues/1037 is fixed.

codemaestro64:
This adds a feature that prevents more than one wallet from using the same seed.

Fixes #132

### DIFF
--- a/multiwallet.go
+++ b/multiwallet.go
@@ -1,6 +1,7 @@
 package dcrlibwallet
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"os"
@@ -282,6 +283,7 @@ func (mw *MultiWallet) CreateNewWallet(walletName, privatePassphrase string, pri
 	if err != nil {
 		return nil, err
 	}
+
 	wallet := &Wallet{
 		Name:                  walletName,
 		CreatedAt:             time.Now(),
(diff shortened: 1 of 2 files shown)


multiwallet.go
```diff
@@ -652,3 +661,14 @@ func (mw *MultiWallet) ChangePrivatePassphraseForWallet(walletID int, oldPrivate
 
 	return nil
 }
+
+// checkForSeedClash checks to see if any wallet in the multi wallet has the same seed as the passed seed
+// returns an error if there is any
+func (mw *MultiWallet) checkForSeedClash(encryptedSeed []byte) error {
+	for _, wal := range mw.wallets {
+		if bytes.Equal(wal.EncryptedSeed, encryptedSeed) {
```
C-ollins, line 669: `EncryptedSeed` cannot be used to compare because the encrypted value is different for every encryption, `EncryptedSeed` is cleared once a wallet is backed up and `EncryptedSeed` is only saved for new wallets.

C-ollins:
Closing this since it's stale and doesn't fix the issue.

PR #139, opened 2020-05-26
spv: use wallet with lowest rescan point to find the lowest rescan point

C-ollins:
This will prevent a bug that'll occur if the wallet with the lowest chain tip does not have the lowest rescan point. The error can be recreated by creating 2 new wallets and trying to sync. I've tested the implementation and it fixes the bug.


raedah:
tested locally and confirmed it resolved the error that was occuring

PR #143, opened 2020-06-16
fix sync progress calculation error

### LINKED ISSUE #95: Sync headers fetch progress ETA calculation error
Below is the code that calculates the progress of an ongoing headers fetch operation.
https://github.com/raedahgroup/dcrlibwallet/blob/d4d67d9b370539a6c0557685bf2319cecaecf5b4/syncnotification.go#L107-L110

The calculation done here is potentially flawed because `fetchedHeadersCount ` which really is `lastFetchedHeaderHeight` may not be the total number of headers fetched so far in THIS sync operation. It may include headers previously fetched.

Recommendation is to determine the actual number of headers fetched so far in THIS sync operation and compare that number against the estimated number of headers left to fetch in the sync operation in order to determine the progress so far.

This calculation error may be the reason for the anomaly previously noted in the dcrlibwallet source code:
https://github.com/raedahgroup/dcrlibwallet/blob/d4d67d9b370539a6c0557685bf2319cecaecf5b4/syncnotification.go#L124-L129

codemaestro64:
This correctly calculates the progress of sync headers fetch operation. It achieves this by comparing the total headers fetched in the current sync operation against estimated number of headers left to be fetched.

Fixes #95

### DIFF
--- a/syncnotification.go
+++ b/syncnotification.go
@@ -59,6 +59,7 @@ func (mw *MultiWallet) fetchHeadersStarted(peerInitialHeight int32) {
 
 	mw.syncData.mu.RLock()
 	headersFetchingStarted := mw.syncData.beginFetchTimeStamp != -1
+	mw.syncData.totalFetchedHeadersCount = 0
 	showLogs := mw.syncData.showLogs
 	mw.syncData.mu.RUnlock()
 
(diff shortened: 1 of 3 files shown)


syncnotification.go
```diff
@@ -59,6 +59,7 @@ func (mw *MultiWallet) fetchHeadersStarted(peerInitialHeight int32) {
 
 	mw.syncData.mu.RLock()
 	headersFetchingStarted := mw.syncData.beginFetchTimeStamp != -1
+	mw.syncData.totalFetchedHeadersCount = 0
```
C-ollins, line 62: Reset this value to zero at the appropriate places.

raedah:
also related https://github.com/planetdecred/dcrandroid/issues/486

PR #153, opened 2020-07-24
Fetch proposals from API

### LINKED ISSUE #152: fetch politeia proposals
A function for fetching all politeia proposals is needed.

This is the API link https://proposals.decred.org/api/v1/proposals/vetted

The preferred response for this would be a blocking response

codemaestro64:
This adds a function that fetches decred proposals from the politea API. 

Fixes #152

### DIFF
--- a/multiwallet.go
+++ b/multiwallet.go
@@ -3,7 +3,6 @@ package dcrlibwallet
 import (
 	"context"
 	"encoding/json"
-	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -36,10 +35,10 @@ type MultiWallet struct {
 	txAndBlockNotificationListeners map[string]TxAndBlockNotificationListener
 	blocksRescanProgressListener    BlocksRescanProgressListener
 
-	httpClient *http.Client
-
 	shuttingDown chan bool
 	cancelFuncs  []context.CancelFunc
+
+	Politea Politea
 }
 
 func NewMultiWallet(rootDir, dbDriver, netType string) (*MultiWallet, error) {
@@ -89,6 +88,7 @@ func NewMultiWallet(rootDir, dbDriver, netType string) (*MultiWallet, error) {
 			syncProgressListeners: make(map[string]SyncProgressListener),
 		},
 		txAndBlockNotificationListeners: make(map[string]TxAndBlockNotificationListener),
+		Politea:                         newPolitea(),
 	}
 
 	// read saved wallets info from db and initialize wallets
@@ -107,9 +107,6 @@ func NewMultiWallet(rootDir, dbDriver, netType string) (*MultiWallet, error) {
 		}
 		mw.wallets[wallet.ID] = wallet
 	}
-
-	// create http client
-	mw.createHTTPClient()
 
 	mw.listenForShutdown()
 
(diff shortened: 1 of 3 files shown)


proposals.go
```diff
@@ -0,0 +1,93 @@
...
+	"net/http"
+	"time"
+)
+
+const (
+	proposalsEndpoint = "https://proposals.decred.org/api/v1/proposals/vetted"
+	policyEndpoint    = "https://proposals.decred.org/api/v1/policy"
+)
+
+func (mw *MultiWallet) createHTTPClient() {
```
C-ollins, line 15: I do not think any of the functions in this file should depend on a multiwallet instance, it should standalone so it can be invoked and managed separately.

proposals.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error fetching proposals from %s: %v", from, err)
+	}
+
+	return result.Proposals, nil
+
+}
+
+// GetAllProposals fetches all the proposals from the API
+func (mw *MultiWallet) GetAllProposals() ([]Proposal, error) {
```
C-ollins, line 64: The return type of this function should be a json string of `[]Proposal` to make it visible after compiling with gomobile.

proposals.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, fmt.Errorf("error fetching politea policy: %v", err)
+	}
+
+	return policy, nil
+}
+
+// GetProposalsChunk gets proposals starting after the proposal with the specified
+// censorship hash. The number of proposals returned is specified in the poltiea
+// policy API endpoint
+func (mw *MultiWallet) GetProposalsChunk(from string) ([]Proposal, error) {
```
C-ollins, line 47: The return type of this function should be a json string of []Proposal to make it visible after compiling with gomobile.

proposals.go
```diff
@@ -0,0 +1,93 @@
...
+	res, err := mw.httpClient.Get(url)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	return json.NewDecoder(res.Body).Decode(responseData)
+}
+
+func (mw *MultiWallet) getPolicy() (*Policy, error) {
```
C-ollins, line 34: Make the function names more descriptive. For example, this function could be `serverPolicy()`.

multiwallet.go
```diff
@@ -37,6 +37,8 @@ type MultiWallet struct {
 
 	shuttingDown chan bool
 	cancelFuncs  []context.CancelFunc
+
+	Politea Politea
```
dreacot, line 41: You need to refractor all instance of "Politea" to "Politeia"

proposals.go
```diff
@@ -0,0 +1,165 @@
...
+	jsonBytes, err := json.Marshal(proposals)
+	if err != nil {
+		return "", fmt.Errorf("error marshalling proposal result to json: %s", err.Error())
+	}
+
+	return string(jsonBytes), nil
+}
+
+// GetAllProposal fetches all vetted proposals from the API
+func (p *Politea) GetAllProposals() (string, error) {
```
dreacot, line 129: This only fetches 20 proposals which I believe is limited by the server policy, it should fetch all proposals

dreacot: This should also follow a similar pattern as fetching transactions, where the offset, limit, filters and newest first flag can be set https://github.com/planetdecred/dcrlibwallet/blob/master/transactions.go#L96

PR #165, opened 2020-10-23
Send custom inputs

song50119:
PR is continues process on [this branch](https://github.com/planetdecred/dcrlibwallet/tree/SendFromCustomInputs) for  UTXO
- Do refactor old code
- Adds utxo.go and move on methods related with UTXO to
- Remove `txhelper/internal/txsizes/size.go` `txhelper/unsignedtx.go` `txhelper/types.go` files since it isn't necessary
- Use `TransactionDestination` type from types.go 
- Refactor (*TxAuthor).ParseOutputsAndChangeDestination
- Refactor (*Wallet).UnspentOutputs
- Refactor (*TxAuthor).constructCustomTransaction
- Refactor calculateChangeScriptSize, calculateMultipleChangeScriptSize


utxo.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		maxSignedSize := txsizes.EstimateSerializeSize(inputScriptSizes, outputs, changeScriptSize)
+		maxRequiredFee := txrules.FeeForSerializeSize(txrules.DefaultRelayFeePerKb, maxSignedSize)
+		changeAmount := totalInputAmount - totalSendAmount - int64(maxRequiredFee)
+
+		if changeAmount < 0 {
+			excessSpending := 0 - changeAmount // equivalent to math.Abs()
+			return nil, 0, fmt.Errorf("total send amount plus tx fee is higher than the total input amount by %s",
+				dcrutil.Amount(excessSpending).String())
```
C-ollins, line 138:
```suggestion
			return nil, 0, errors.New(ErrInsufficientBalance)
```

Since you're sending using custom inputs, I assume you'll handle this error in the frontend as insufficient inputs, not insufficient balance.

utxo.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		if changeAmount != 0 && !txrules.IsDustAmount(dcrutil.Amount(changeAmount),
+			changeScriptSize, txrules.DefaultRelayFeePerKb) {
```
C-ollins, line 142:
```suggestion
		if changeAmount != 0 && !txrules.IsDustAmount(dcrutil.Amount(changeAmount), changeScriptSize, txrules.DefaultRelayFeePerKb) {
```
A line break isn't needed here.

txauthor.go
```diff
@@ -113,6 +136,46 @@ func (tx *TxAuthor) EstimateMaxSendAmount() (*Amount, error) {
 	}, nil
 }
 
+func (tx *TxAuthor) UseInputs(utxoKeys []string) error {
```
C-ollins, line 139: What's utxo key? Do you mean hash?

txauthor.go
```diff
@@ -113,6 +136,46 @@ func (tx *TxAuthor) EstimateMaxSendAmount() (*Amount, error) {
...
+	}
+
+	retrieveAccountUtxo := func(utxoKey string) *UnspentOutput {
+		for _, accountUtxo := range accountUtxos {
+			if accountUtxo.OutputKey == utxoKey {
+				return accountUtxo
+			}
+		}
+		return nil
+	}
```
C-ollins, line 155: Instead of getting all utxos and looping through the slice each time you want to find a utxo, why don't you add a function to find utxo by transaction outpoint?.   
You can create output using the hash and index(you can get that by splitting `utxoKey` by ":") and get a single utxo the way it's done here: https://github.com/planetdecred/dcrlibwallet/blob/f3785c80f2f344a10f07715c3bac9ac4ede37712/accounts.go#L144-L147

C-ollins (changes requested):

utxo.go
song50119: Should I add more new of errors type, like `ErrInsufficientInputsBalance`?

txauthor.go
```diff
@@ -64,6 +69,31 @@ func (tx *TxAuthor) SendDestination(atIndex int) *TransactionDestination {
 	return &tx.destinations[atIndex]
 }
 
+func (tx *TxAuthor) AddChangeDestination(address string, atomAmount int64) int {
```
C-ollins, line 72: - Why does change destination have an amount parameter?
- How will the user know the change amount without knowing the fee?
- What happens if the change amount is lower than the actual change and will this result in paying a higher fee?

utxo.go
```diff
@@ -0,0 +1,192 @@
...
+	inputScriptSizes := make([]int, len(inputs))
+	inputScripts := make([][]byte, len(inputs))
+	for i, input := range inputs {
+		totalInputAmount += input.ValueIn
+		inputScriptSizes[i] = txsizes.RedeemP2PKHSigScriptSize
+		inputScripts[i] = input.SignatureScript
+	}
+
+	var changeScriptSize int
+	if maxAmountRecipientAddress != "" {
```
C-ollins, line 125: This will always evaluate to true because you assigned it here:
https://github.com/planetdecred/dcrlibwallet/blob/728a131272971ba04eb89b47736ff84d99a24fdb/utxo.go#L107-L113

I'd suggest you replace:
https://github.com/planetdecred/dcrlibwallet/blob/728a131272971ba04eb89b47736ff84d99a24fdb/utxo.go#L107

With:`if maxAmountRecipientAddress == "" && len(changeDestinations) == 0 {` to make this logic work.

utxo.go
```diff
@@ -0,0 +1,192 @@
...
+			// randomize the change output that was just added
+			changeOutputIndex := len(outputs) - 1
+			txauthor.RandomizeOutputPosition(outputs, changeOutputIndex)
+		}
+
+		if totalChangeAmount > changeAmount {
+			return nil, fmt.Errorf("total amount allocated to change addresses (%s) is higher than"+
+				" actual change amount for transaction (%s)", dcrutil.Amount(totalChangeAmount).String(),
+				dcrutil.Amount(changeAmount).String())
+		}
```
C-ollins, line 175: Put this block in an else statement.
```suggestion
		} else if len(changeDestinations) > 0 {

			var totalChangeAmount int64
			for _, changeDestination := range changeDestinations {
				changeOutput, err := txhelper.MakeTxOutput(changeDestination.Address,
					changeDestination.AtomAmount, tx.sourceWallet.chainParams)
				if err != nil {
					return nil, fmt.Errorf("change address error: %v", err)
				}

				totalChangeAmount += changeOutput.Value
				outputs = append(outputs, changeOutput)

				// randomize the change output that was just added
				changeOutputIndex := len(outputs) - 1
				txauthor.RandomizeOutputPosition(outputs, changeOutputIndex)
			}

			if totalChangeAmount > changeAmount {
				return nil, fmt.Errorf("total amount allocated to change addresses (%s) is higher than"+
					" actual change amount for transaction (%s)", dcrutil.Amount(totalChangeAmount).String(),
					dcrutil.Amount(changeAmount).String())
			}
		}
```

utxo.go
```diff
@@ -0,0 +1,192 @@
...
+	} else {
+		maxSignedSize = txsizes.EstimateSerializeSize(inputScriptSizes, outputs, 0)
```
C-ollins, line 177: Why is this done here? You already calculated the max signed size with the correct change script size and input script sizes here:
https://github.com/planetdecred/dcrlibwallet/blob/728a131272971ba04eb89b47736ff84d99a24fdb/utxo.go#L134

C-ollins (changes requested):

C-ollins (changes requested):

utxo.go
```diff
@@ -0,0 +1,193 @@
...
+// Returns an error if more than 1 max amount recipients identified or
+// if any other error is encountered while processing the addresses and amounts.
+func (tx *TxAuthor) ParseOutputsAndChangeDestination(txDestinations []TransactionDestination) ([]*wire.TxOut, int64, string, error) {
+	var outputs = make([]*wire.TxOut, 0)
+	var totalSendAmount int64
+	var maxAmountRecipientAddress string
+
+	for _, destination := range txDestinations {
+		// validate the amount to send to this destination address
+		if !destination.SendMax && (destination.AtomAmount <= 0 || destination.AtomAmount > dcrutil.MaxAmount) {
```
C-ollins, line 53: Don't do amount validation here, do it when adding the send destination [here](https://github.com/planetdecred/dcrlibwallet/blob/f1488744385e850f0f411de7508edd0598e81879/txauthor.go#L39) and [here](https://github.com/planetdecred/dcrlibwallet/blob/f1488744385e850f0f411de7508edd0598e81879/txauthor.go#L54)

utxo.go
```diff
@@ -0,0 +1,193 @@
...
+		if maxAmountRecipientAddress != "" {
+			outputs, err = tx.changeOutput(changeAmount, maxAmountRecipientAddress, outputs)
+			if err != nil {
+				return nil, fmt.Errorf("change address error: %v", err)
+			}
+			totalChangeAmount = changeAmount
+		} else if len(changeDestinations) > 0 {
+			changeDestinations[0].AtomAmount = changeAmount
+			for _, changeDestination := range changeDestinations {
+				newOutputs, err := tx.changeOutput(changeDestination.AtomAmount, changeDestination.Address, outputs)
```
C-ollins, line 156: `changeDestination.AtomAmount` will always be 0, remember that amount parameter was removed for change destination.
https://github.com/planetdecred/dcrlibwallet/blob/f1488744385e850f0f411de7508edd0598e81879/txauthor.go#L72

utxo.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		var totalChangeAmount int64
+		if maxAmountRecipientAddress != "" {
+			outputs, err = tx.changeOutput(changeAmount, maxAmountRecipientAddress, outputs)
+			if err != nil {
+				return nil, fmt.Errorf("change address error: %v", err)
+			}
+			totalChangeAmount = changeAmount
+		} else if len(changeDestinations) > 0 {
+			changeDestinations[0].AtomAmount = changeAmount
```
C-ollins, line 154: Why is the entire change amount assigned to the first change destination?, this means there'll be no change left for the remaining change destinations.

C-ollins (changes requested):

utxo.go
song50119: On the mobile or wallet if developer is forget to check errors at `AddSendDestination` or `UpdateSendDestination`, we should validate again here, right?

C-ollins (approved):

PR #167, opened 2020-12-18
Implement cfilters sync notification and fix badger ErrTxnTooBig error

C-ollins:
Cfilters progress notification never came in for previous dcrlibwallet/dcrwallet versions so a notification handler wasn't implemented. An upgrade(for existing synced wallets) to dcrwallet's wallet DB version from version [13](https://github.com/planetdecred/dcrlibwallet/releases/tag/v1.5.2) to [15](https://github.com/planetdecred/dcrlibwallet/commit/d860a56b0b69c578cbcee791ad89b5261fa6d786) will cause previously saved cfilters to be deleted, therefore cfilters fetch for all downloaded blocks will be needed once and during the next sync. The cfiters fetch happens once and will not be received in subsequent syncs once it's completed. This PR implements the handlers for the cfilters fetch notification and the sync estimation code is adjusted to account for cfilters fetch time.  
This PR also fixes a memory error `ErrTxnTooBig` from badger that occurs during dcrwallet's wallet DB upgrade. During the upgrade to version 15, all saved cfilters are [deleted](https://github.com/decred/dcrwallet/blob/master/wallet/udb/upgrades.go#L1279) and this means the database is tasked to delete over 500K entries(for a synced testnet wallet) using one transaction. Due to memory limitations on android, badger is set to use 64MB max for it's memtable(each db transaction gets [15%](https://github.com/dgraph-io/badger/blob/8dbc982d4975a76318b42525adc670b80b491f63/db.go#L195)) and that allocated memory gets maxed out during the wallet DB upgrade process which causes badger to return `ErrTxnTooBig`. The error is fixed by committing the current DB transaction whenever it gets full and creates a new one.


sync.go
```diff
@@ -39,38 +39,48 @@ type syncData struct {
...
 
 const (
 	InvalidSyncStage          = -1
-	HeadersFetchSyncStage     = 0
-	AddressDiscoverySyncStage = 1
-	HeadersRescanSyncStage    = 2
+	CFiltersFetchSyncStage    = 0
+	HeadersFetchSyncStage     = 1
+	AddressDiscoverySyncStage = 2
+	HeadersRescanSyncStage    = 3
```
oshorefueled, line 59: you can use the iota enum type for successive integer increments

syncnotification.go
```diff
@@ -50,6 +50,104 @@ func (mw *MultiWallet) handlePeerCountUpdate(peerCount int32) {
 	}
 }
 
+// Fetch CFilters Callbacks
+
```
oshorefueled, line 54: Comment can be more descriptive

oshorefueled (changes requested):

sync.go
C-ollins: iota enum would not be visible when compiled with gomobile which is why it's written this way.

syncnotification.go
C-ollins: That comment is used to mark the beginning of cfilters callbacks. Similar comments are used to mark other sync stages;
https://github.com/planetdecred/dcrlibwallet/blob/38838d9adce251fe25d679fdc73c77ec87ac5898/syncnotification.go#L151 https://github.com/planetdecred/dcrlibwallet/blob/38838d9adce251fe25d679fdc73c77ec87ac5898/syncnotification.go#L310
https://github.com/planetdecred/dcrlibwallet/blob/38838d9adce251fe25d679fdc73c77ec87ac5898/syncnotification.go#L455

oshorefueled (approved):

PR #178, opened 2021-02-11
Allow dcrlibwallet to use an external logger

oluwandabira:
Currently dcrlibwallet uses its own internal log system which is fine but as dcrlibwallet is mostly used as a package and not a standalone program this doesn't allow programs that uses dcrlibwallet to control the logging (save for setting the log levels).

This pull request will allow external programs to use their own log systems in place of this internal system allowing greater control of logging output. 

For example, in [godcr](https://github.com/planetdecred/godcr) this will allow a cross platform way for the application to display a page showing the log output.

This doesn't replace the current system in any capacity, dcrlibwallet will still use its own system if no logger is used.

### DIFF
--- a/log.go
+++ b/log.go
@@ -98,16 +98,23 @@ func initLogRotator(logFile string) error {
 	return nil
 }
 
+// UseLoggers sets the subsystem logs to use the provided loggers.
+func UseLoggers(main, loaderLog, walletLog, tkbyLog,
+	syncLog, cmgrLog, amgrLog slog.Logger) {
+	log = main
+	loader.UseLogger(loaderLog)
+	wallet.UseLogger(walletLog)
+	udb.UseLogger(walletLog)
+	ticketbuyer.UseLogger(tkbyLog)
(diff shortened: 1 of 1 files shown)


log.go
```diff
@@ -98,6 +98,18 @@ func initLogRotator(logFile string) error {
...
+func UseLogger(logger slog.Logger) {
+	log = logger
+	loader.UseLogger(logger)
+	wallet.UseLogger(logger)
+	udb.UseLogger(logger)
+	ticketbuyer.UseLogger(logger)
+	spv.UseLogger(logger)
+	p2p.UseLogger(logger)
+	connmgr.UseLogger(logger)
+	addrmgr.UseLogger(logger)
```
C-ollins, line 110: Using the same logger for every package will assign whatever tag you give dcrlibwallet to every package. Can you register a logger for each package from godcr?

C-ollins (changes requested):

C-ollins (approved):

PR #188, opened 2021-05-02
Ability to config VSP

song50119:
- Add new methods into vspd.go to be able to save, update, delete VSP

### DIFF
--- a/vspd.go
+++ b/vspd.go
@@ -51,6 +51,60 @@ func (mw *MultiWallet) NewVSPD(baseURL string, walletID int, sourceAccountNumber
 	}
 }
 
+func (mw *MultiWallet) SaveVSPConfigValue(url string) (error, *GetVspInfoResponse) {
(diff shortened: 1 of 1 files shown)


vspd.go
```diff
@@ -51,6 +51,60 @@ func (mw *MultiWallet) NewVSPD(baseURL string, walletID int, sourceAccountNumber
 	}
 }
 
+func (mw *MultiWallet) SaveVSPConfigValue(url string) (error, *GetVspInfoResponse) {
+	v := &VSPD{baseURL: url, httpClient: new(http.Client)}
+	info, err := v.GetInfo()
```
C-ollins, line 56: Why do we need to save vsp host config? That sounds like data that should be fetched on demand because the values could change anytime.

song50119:
This can be done by use method `SaveUserConfigValue ` with exported VSPHostConfigKey will be use

PR #204, opened 2021-07-09
Add function to get time till next ticket price change

### LINKED ISSUE #203: duration till next ticket price
There should be a method on dcrlibwallet that returns the duration till next ticket price is determined.

song50119:
Resolves #203


ticket.go
```diff
@@ -376,3 +377,50 @@ func CallVSPTicketInfoAPI(vspHost, pubKeyAddr string) (ticketPurchaseInfo *VSPTi
...
+		str += fmt.Sprintf("%dm ", mins)
+	}
+	if allsecs > 0 {
+		str += fmt.Sprintf("%ds ", allsecs)
+	}
+	return str
+}
+
+// NextTicketPriceRemaining returns the remaning time of a ticket for the next block
+func (mw *MultiWallet) NextTicketPriceRemaining(netType string) (string, error) {
```
C-ollins, line 415: Return the remaining time in seconds `int64` instead of a string.

C-ollins (changes requested):

ticket.go
C-ollins: Remove the `netType` parameter and use `mw.chainParams`.

C-ollins (approved):

PR #208, opened 2021-09-23
add function to get the last synced timestamp for politeia

dreacot:
This PR adds a function to track when last proposals were synced from the politeia server. It would be helpful to display the last update time in the clients apps


politeia_sync.go
```diff
@@ -62,11 +62,17 @@ func (p *Politeia) Sync() error {
...
 		log.Info("Politeia sync: update complete")
+		p.lastSyncedTimestamp = time.Now()
 		p.publishSynced()
 		return nil
 	}
 }
 
+func (p *Politeia) GetLastSyncedTimeStamp() string {
+	currentTime := time.Now()
+	return TimeElapsed(currentTime, p.lastSyncedTimestamp, false)
```
Sirmorrison, line 73: This should be sent as a timestamp instead.

utils.go
```diff
@@ -322,3 +322,73 @@ func initWalletLoader(chainParams *chaincfg.Params, walletDataDir, walletDbDrive
...
 }
+
+func s(x float64) string {
+	if int(x) == 1 {
+		return ""
+	}
+	return "s"
+}
+
+func TimeElapsed(now time.Time, then time.Time, full bool) string {
```
Sirmorrison, line 333: I should be able to decide the format i want the time to be example `m`, `min(s)`, `minute(s)`

Sirmorrison (changes requested):

utils.go
Sirmorrison: So this is a Dcrlibwallet method used on godcr with timestamp. https://github.com/planetdecred/godcr/blob/d9f384c76ab1068b3f01b90810207dcd1c104244/wallet/commands.go#L117

This and many more methods are in the dcrlibwallet utils.go file.

Having this, it will help eliminate repeating same method on godcr and the other platform.
especially this as the logic is the same.
https://github.com/planetdecred/godcr/blob/d9f384c76ab1068b3f01b90810207dcd1c104244/ui/page/components/components.go#L439

dreacot: oh okay, noted, thanks for clearing it up 👍🏼

utils.go
```diff
@@ -322,3 +322,121 @@ func initWalletLoader(chainParams *chaincfg.Params, walletDataDir, walletDbDrive
 
 	return walletLoader
 }
+
+func s(x float64) string {
```
C-ollins, line 326: Give this an appropriate name.

politeia.go
```diff
@@ -18,6 +18,7 @@ type Politeia struct {
 	ctx                     context.Context
 	cancelSync              context.CancelFunc
 	client                  *politeiaClient
+	lastSyncedTimestamp     int64
```
C-ollins, line 21: This value should be saved to the db. The last synced time will be `0` if the app hasn't done a successful politeia sync.

C-ollins (dismissed):

politeia_sync.go
```diff
@@ -62,11 +62,17 @@ func (p *Politeia) Sync() error {
 		}
 
 		log.Info("Politeia sync: update complete")
+		now := time.Now()
+		p.lastSyncedTimestamp = now.Unix()
```
Sirmorrison, line 66:
```suggestion
		p.lastSyncedTimestamp = time.Now().Unix()
```

politeia_sync.go
```diff
@@ -62,11 +62,17 @@ func (p *Politeia) Sync() error {
 		}
 
 		log.Info("Politeia sync: update complete")
+		p.lastSyncedTimestamp = time.Now().Unix()
+		p.saveLastSyncedTimestamp(p.lastSyncedTimestamp)
```
Sirmorrison, line 66: I dont think `p.lastSyncedTimestamp` is needed since it is not used outside this method.
```suggestion
		p.saveLastSyncedTimestamp(time.Now().Unix())
```

politeia.go
Sirmorrison, line 21: this is redundant.

utils.go
```diff
@@ -322,3 +322,123 @@ func initWalletLoader(chainParams *chaincfg.Params, walletDataDir, walletDbDrive
...
+
+	if week > 0 {
+		if abbreviationFormat == "long" {
+			parts = append(parts, strconv.Itoa(int(week))+" week"+makePlural(week))
+		} else if abbreviationFormat == "short" {
+			parts = append(parts, strconv.Itoa(int(week))+" wk"+makePlural(week))
+		} else if abbreviationFormat == "shortest" {
+			parts = append(parts, strconv.Itoa(int(week))+" w")
+		}
+	}
```
Sirmorrison, line 388: We can declare the repeating `long`, `short`, `shortest` as variables, and make use of the variables instead.

Sirmorrison (changes requested):

Sirmorrison (approved):

PR #210, opened 2021-10-02
Add dexdcr package to implement dex wallet functionality.

itswisdomagain:
The [default dcrdex backend driver for Decred wallets](https://github.com/decred/dcrdex/blob/77f324791a85f54c79189c48ee03766d97d67749/client/asset/dcr/dcr.go#L335-L341) uses json-rpc to perform wallet operations.

This adds a dexdcr package to provide an alternative backend driver (`dexdcr.Driver`) for standalone Decred SPV wallets (no json-rpc required). The `dexdcr.Driver` instead requires an instance of a loaded `*decred.org/dcrwallet/v2/wallet.Wallet` which it uses to perform wallet operations.

Dex client apps that want to use this alternative backend would register the `dexdcr.Driver` instead of the default one provided by the dcrdex package. This driver however currently only supports `*decred.org/dcrwallet/v2/wallet.Wallet` that are connected to a supported SPV syncer viz `*decred.org/dcrwallet/v2/spv.Syncer` or `*github.com/planetdecred/dcrlibwallet/spv.WalletBackend`.

The `dexdcr.Driver` uses an instance of the `dexdcr.SpvWallet` struct to provide wallet functionality. The `dexdcr.SpvWallet` works very similarly to [the json-rpc server](https://github.com/decred/dcrwallet/blob/4f3b2d70ea259632fd8328fc010ddf30ec552aa2/internal/rpc/jsonrpc/methods.go#L81-L194) that handles wallet operations for the default Decred wallet driver used by the dcrdex project. This ensures that the newly added driver will behave very similarly to the default driver.


itswisdomagain:
Note to reviewers:
As explained in the PR description above, it is vital that the `SpvWallet` struct satisfies the `` in a way that closely matches the wallet json-rpc server. It's imperative (now and subsequently) to compare how wallet operations are performed here against how they are done in the json-rpc server, especially since every wallet operation that is implemented here has a json-rpc counterpart.

The json-rpc implementations can be found here: https://github.com/decred/dcrwallet/blob/4f3b2d70ea259632fd8328fc010ddf30ec552aa2/internal/rpc/jsonrpc/methods.go#L81-L194

itswisdomagain:
This is being tested from https://github.com/planetdecred/godcr/pull/637, so a couple more commits could be added as issues are uncovered. But review can happen simultaneously as well.

C-ollins (commented):
LGTM

PR #212, opened 2021-11-12
Add dex client features to MultiWallet

itswisdomagain:
Add MultiWallet methods to provide core features for starting and using a DEX client. Builds on the `dexdcr` package that allows using a custom dcr wallet rather than using an rpc connection to a running dcrwallet instance.

In draft while it's tested from https://github.com/planetdecred/godcr/pull/676 and subsequent/follow-up PRs.


dexclient.go
```diff
@@ -0,0 +1,363 @@
...
+// the DEX client database.
+func (d *DexClient) Reset() bool {
+	shutdownOk := d.shutdown(false)
+	if !shutdownOk {
+		return false
+	}
+
+	err := os.RemoveAll(d.dexDataDir)
+	if err != nil {
+		d.log.Warnf("DEX client reset failed: erroring deleting DEX db: %v", err)
```
Sirmorrison, line 172: There is a typo here `erroring`

Sirmorrison (commented):
PR looks good to me just the typo

PR #214, opened 2021-11-30
update decred.org/dcrdex dep, remove dcrlibwallet dep from dexdcr

itswisdomagain:
Update the dcrdex dep to use recent changes relating to dcr spv wallet support. Also unrequire dcrlibwallet from the dexdcr package to prevent import cycle errors that would occur when requiring dexdcr from dcrlibwallet.


dexdcr/wallet.go
```diff
@@ -259,55 +259,64 @@ func (spvw *SpvWallet) LockUnspent(ctx context.Context, unlock bool, ops []*wire
...
 
-	bestHash, bestHeight := spvw.wallet.MainChainTip(ctx)
-	var confirmations int64
-	if utxo.Block.Height != -1 {
-		confirmations = int64(confirms(utxo.Block.Height, bestHeight))
+		return &dcr.TxOutput{
+			TxOut: &wire.TxOut{
+				Value:    int64(utxo.Amount),
+				PkScript: utxo.PkScript,
+				// Version: 0, // TODO: dcrwallet json-rpc doesn't set this also, but should be
```
dreacot, line 310: This would be done in another PR right?

itswisdomagain: It'll be done when dcrwallet does it too. The idea is to closely simulate the dcrwallet json-rpc methods as that's what the main dex project uses for dcr wallet operations. It's something that should ideally be done but not sure there are any plans towards that yet. It's safe as is for now though.

dreacot (approved):

PR #215, opened 2021-12-06
Add auto ticket buyer

Sirmorrison:
This PR implemented the Ticket buyer functionality in drclibwallet.
Migrated all VSP related methods from godcr to dcrlibwallet
- [x]  Fix cancel ticket buyer bug.
- [x] fix balance to maintain bug.


vsp.go
```diff
@@ -110,6 +114,240 @@ func (v *VSP) PurchaseTickets(ticketCount, expiryBlocks int32, passphrase []byte
...
+				// being mined in the interval after the next.
+				expiry = nextIntervalStart
+				if height+1 == nextIntervalStart {
+					expiry += intervalSize
+				}
+			}
+
+			cancelCtx, cancel := context.WithCancel(ctx)
+			cancels = append(cancels, cancel)
+			go func() {
```
itswisdomagain, line 228: I notice here that dcrwallet appears to have support for buying more than 1 ticket per block, but we don't have that here. I also see that the dcrwallet ticket buyer has some relationship with the mixer but I haven't dived deep to see the nature of the relationship. I did notice though that dcrlibwallet's account mixer seems to set up a ticket buyer as well and wonder why we aren't using that here.

https://github.com/planetdecred/dcrlibwallet/blob/1cdba274e194b8c8090b82026bfbf970e4668133/account_mixer.go#L164
https://github.com/planetdecred/dcrlibwallet/blob/1cdba274e194b8c8090b82026bfbf970e4668133/account_mixer.go#L200-L208
https://github.com/planetdecred/dcrlibwallet/blob/1cdba274e194b8c8090b82026bfbf970e4668133/account_mixer.go#L223

itswisdomagain (changes requested):
Stopping my review for not to discuss some differences noticed between this PR's implementation of the ticket buyer and dcrwallet's. Comments below.

vsp.go
Sirmorrison: The limits to the number of tickets to be purchase is currently set to `TODO` the feature to increase the limits per blocks will be added later on. It is currently set to default `1` due to the meeting had with raedah some few day ago.

Ticket mixer and ticket buyer for `dcrwallet` share same functionally but runs different process depending on the flag set. `c.MixChange` and `c.BuyTickets`.
The reason i am not using the dcrwallet ticket buyer dirrectly is the `c.VSP` config param. it is not letting me set the value to dcrlibwallet `vspClient`.

itswisdomagain: Interesting point there. I have a diff in the works that may allow you set the c.VSP param. Will update later.

vsp.go
```diff
@@ -110,6 +114,240 @@ func (v *VSP) PurchaseTickets(ticketCount, expiryBlocks int32, passphrase []byte
...
+		}
+
+		v.w.cancelAccountMixer = nil
+	}()
+
+	return nil
+}
+
+func (v *VSP) ticketBuyerInstance(ctx context.Context, balanceToMaintain int64, passphrase []byte) error {
+	if len(passphrase) > 0 {
```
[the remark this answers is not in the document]
Sirmorrison: Directly from the endpoint yes. But from the UI, not likely.

vsp.go
```diff
@@ -110,6 +114,240 @@ func (v *VSP) PurchaseTickets(ticketCount, expiryBlocks int32, passphrase []byte
...
+				intervalSize := int32(w.ChainParams().StakeDiffWindowSize)
+				currentInterval := height / intervalSize
+				nextIntervalStart = (currentInterval + 1) * intervalSize
+
+				// Skip this purchase when no more tickets may be purchased in the interval and
+				// the next sdiff is unknown.  The earliest any ticket may be mined is two
+				// blocks from now, with the next block containing the split transaction
+				// that the ticket purchase spends.
+				if height+2 == nextIntervalStart {
+					log.Info("Skipping purchase: next sdiff interval starts soon")
```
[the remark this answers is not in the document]
Sirmorrison: was testing with this let me revert.

types.go
```diff
@@ -499,4 +499,13 @@ type PendingFee struct {
...
 }
 
+type VSPInfo struct {
+	Host string
+	Info *VspInfoResponse
+}
+
+type VSPList struct {
+	List []*VSPInfo
+}
```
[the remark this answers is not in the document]
Sirmorrison: This would not work as expected as maps do not support indexing, and i would need to convert back to a slice in the target application (godcr).
Also, `Host` is not a property in `*VspInfoResponse` hence having to get the key `Host` for use later on is a challenge.

vsp.go
```diff
@@ -110,6 +114,240 @@ func (v *VSP) PurchaseTickets(ticketCount, expiryBlocks int32, passphrase []byte
...
+	if balanceToMaintain < 0 {
+		return errors.New("Negative balance to maintain given")
+	}
+
+	go func() {
+		log.Info("Running ticket buyer")
+		ctx, cancel := context.WithCancel(v.ctx)
+		v.w.cancelAutoTicketBuyer = cancel
+
+		err := v.ticketBuyerInstance(ctx, balanceToMaintain, passphrase)
```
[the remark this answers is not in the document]
Sirmorrison: Adding the lock causes the buying process to fail with the following log.
```2022-01-21 13:30:25.769 [ERR] DLWL: BUY: wallet.PurchaseTickets: wallet locked:: individualSplit: default
2022-01-21 13:30:25.769 [ERR] DLWL: Ticket purchasing failed: wallet.PurchaseTickets: wallet locked:: individualSplit: default```

File: vsp.go [hunk 6]
```diff
@@ -110,6 +114,240 @@ func (v *VSP) PurchaseTickets(ticketCount, expiryBlocks int32, passphrase []byte
...
 
+func (v *VSP) StartTicketBuyer(balanceToMaintain int64, passphrase []byte) error {
+	if balanceToMaintain < 0 {
+		return errors.New("Negative balance to maintain given")
+	}
+
+	go func() {
+		log.Info("Running ticket buyer")
+		ctx, cancel := context.WithCancel(v.ctx)
+		v.w.cancelAutoTicketBuyer = cancel
```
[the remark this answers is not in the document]
Sirmorrison: > Why is this a property of the wallet rather than the vsp instance? The vsp instance starts the tb via this `vsp.StartTicketBuyer` method, makes sense for it to also be the one to stop it.

This was made so, so that we can always check the state of the VSP without having to start/initialize a new VSP for every instance were is it to be used.

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	if len(passphrase) > 0 {
+		err := vsp.w.UnlockWallet(passphrase)
+		if err != nil {
+			return translateError(err)
+		}
+	}
+
+	if vsp.mwRef.IsAutoTicketsPurchaseActive(vsp.w.ID) {
+		return errors.New("Ticket buyer already running")
+	}
```
itswisdomagain, line 134: Make this check before unlocking the wallet.

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			log.Errorf("Ticket buyer instance errored: %v", err)
+		}
+
+		vsp.w.cancelAutoTicketBuyer = nil
+	}()
+
+	return nil
+}
+
+func (vsp *VSP) ticketBuyer(ctx context.Context, balanceToMaintain int64) error {
```
itswisdomagain, line 152: When the name of a method/function is a noun, it is generally interpreted to be a getter method/function. E.g. `VSP()` should get/return a vsp, `WalletWithID()` should return a wallet etc. Methods/functions that perform tasks and return a response but not an object should be verbs. E.g. `Sync()`, `UnlockWallet()`. This should be `startTicketBuyer` or something along that line.

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	}
+
+	if vsp.mwRef.IsAutoTicketsPurchaseActive(vsp.w.ID) {
+		return errors.New("Ticket buyer already running")
+	}
+
+	go func() {
+		log.Info("Running ticket buyer")
+
+		ctx, cancel := vsp.mwRef.contextWithShutdownCancel()
```
itswisdomagain, line 139:
```suggestion
		ctx, cancel := vsp.w.shutdownContextWithCancel()
```
Don't need `mwRef`.

vsp.go
itswisdomagain: Instead of adding a reference to mw, make this a mw or wallet method. The method can then make this check at the  beginning, create a vsp instance and call the private `vsp.startTicketBuyer` method.

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			return translateError(err)
+		}
+	}
+
+	if vsp.mwRef.IsAutoTicketsPurchaseActive(vsp.w.ID) {
+		return errors.New("Ticket buyer already running")
+	}
+
+	go func() {
+		log.Info("Running ticket buyer")
```
itswisdomagain, line 137: I think this log message should include the wallet ID and/or name.

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			w := vsp.w.Internal()
+
+			// Don't perform any actions while transactions are not synced through
+			// the tip block.
+			rp, err := w.RescanPoint(ctx)
+			if err != nil {
+				return err
+			}
+			if rp != nil {
+				log.Debugf("Skipping autobuyer actions: transactions are not synced")
```
itswisdomagain, line 189: Is it possible for logs to include the wallet ID/name so users running ticket buyers on multiple wallets can identify which wallet is producing which log message?

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+						outerCancel()
+					}
+				}
+			}()
+		}
+	}
+}
+
+// buyTicket calculate the number of tickets before purchase
+func (vsp *VSP) buyTicket(ctx context.Context, tip *wire.BlockHeader, expiry int32, maintain int64) error {
```
itswisdomagain, line 259: If this method is allowed to buy more than 1 ticket, the following should work. Format the comment to max 80 chars per line.

```suggestion
// buyTickets purchases one or more tickets depending on the spendable balance of the selected account and the specified minimum balance to maintain.
func (vsp *VSP) buyTickets(ctx context.Context, tip *wire.BlockHeader, expiry int32, maintain int64) error {
```

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+
+	wal := vsp.w
+
+	networkBackend, err := wal.Internal().NetworkBackend()
+	if err != nil {
+		return err
+	}
+
+	// Determine how many tickets to buy
+	bal, err := wal.GetAccountBalance(int32(vsp.purchaseAccount))
```
itswisdomagain, line 271:
```suggestion
	// Get the account balance to determine how many tickets to buy.
	bal, err := wal.GetAccountBalance(int32(vsp.purchaseAccount))
```

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	tix, err := vsp.w.Internal().PurchaseTickets(ctx, networkBackend, request)
+	if tix != nil {
+		for _, hash := range tix.TicketHashes {
+			log.Infof("Purchased ticket %v at stake difficulty %v", hash, sdiff)
+		}
+	}
+
+	if err != nil {
+		log.Errorf("BUY: %v", err)
+	}
```
itswisdomagain, line 323: The returned error is logged by the caller when necessary. No need to log here.

vsp.go
```diff
@@ -440,18 +682,187 @@ func (v *VSP) GetFeeAddress(ctx context.Context, ticketHash chainhash.Hash) (dcr
...
+func (mw *MultiWallet) TicketBuyerConfigIsSet() bool {
+	return mw.ReadBoolConfigValueForKey(TicketBuyerConfigSet, false)
+}
+
+func (mw *MultiWallet) ClearTicketBuyerConfig() {
+	mw.SetLongConfigValueForKey(TicketBuyerATMConfigKey, -1)
+	mw.SetIntConfigValueForKey(TicketBuyerWalletConfigKey, -1)
+	mw.SetInt32ConfigValueForKey(TicketBuyerAccountConfigKey, -1)
+	mw.SetStringConfigValueForKey(TicketBuyerVSPHostConfigKey, "")
+	mw.SetBoolConfigValueForKey(TicketBuyerConfigSet, false)
```
itswisdomagain, line 827: It seems the only time `TicketBuyerConfigSet` is false is also when `TicketBuyerVSPHost` is cleared. Can do away with the extra config field and just check if the host is empty.

vsp.go
```diff
@@ -440,18 +682,187 @@ func (v *VSP) GetFeeAddress(ctx context.Context, ticketHash chainhash.Hash) (dcr
...
+}
+
+func (wallet *Wallet) listenForTicketBuyerShutdown() {
```
itswisdomagain, line 830: Don't see where this method is used and doesn't seem to serve any purpose. Want to remove?

itswisdomagain (changes requested):

vsp.go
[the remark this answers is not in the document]
itswisdomagain: It seems I was wrong. I do see a `defer vsp.w.LockWallet()` in the `vsp.PurchaseTickets` method. The `LockWallet` should probably check though that ticket buyer isn't running before actually locking the wallet so the ticket buyer doesn't get affected if the wallet gets locked by another process. Alternatively, you can have the ticket buyer keep the passphrase in memory, check if the wallet is locked and then unlock it each time before buying tickets. Wallet does this.

vsp.go
```diff
@@ -60,62 +66,298 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	request := &w.PurchaseTicketsRequest{
+		Count:                buy,
+		SourceAccount:        vsp.purchaseAccount,
+		Expiry:               expiry,
+		MinConf:              DefaultRequiredConfirmations,
+		VSPFeeProcess:        vsp.PoolFee,
+		VSPFeePaymentProcess: vsp.ProcessFee,
+	}
+
+	tix, err := vsp.w.Internal().PurchaseTickets(ctx, networkBackend, request)
```
itswisdomagain, line 314: It's worth noting that this wallet method will first mix funds before purchasing a ticket if cspp parameters are provided in `request`. If the user has enabled mixing, it would be vital to add the cssp params here so wallet can do some mixing before purchasing tickets. This is necessary even if purchasing from the mixed account because unless the mixed account has utxos with the exact ticket amount, one or more utxos will be used to create a new utxo with the exact ticket amount before the ticket will be purchased. This means that 1 or more mixed outputs may be combined into one transaction, and used to create an output of the same value as the ticket price, which will then be used to purchase the ticket. This risks undoing the results of the mixing in the first place.

itswisdomagain: Perhaps this can be added to #225 after this goes in because #225 already makes a note to do this for regular ticket purchases, i.e. not automated ticket purchases.

itswisdomagain (changes requested):

vsp.go
Sirmorrison: It is not possible to run ticket buyer on multiple wallets at same time.. you will need to cancel and start a new process with the new wallet.

vsp.go
```diff
@@ -57,65 +61,309 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+				// being mined in the interval after the next.
+				expiry = nextIntervalStart
+				if height+1 == nextIntervalStart {
+					expiry += intervalSize
+				}
+			}
+
+			cancelCtx, cancel := context.WithCancel(ctx)
+			cancels = append(cancels, cancel)
+			go func() {
```
itswisdomagain, line 235: Based on recent conversation regarding this feature, this will need to more closely resemble dcrwallet's pattern here:
https://github.com/decred/dcrwallet/blob/9a8688e60b08b68b5952bb717122669cbb3e0989/ticketbuyer/tb.go#L199-L201

That should look like this in our case:
```
			// multiple is the max amount of tickets that can be bought based on available funds
			// calculate it here and start separate purchase goroutines for as many tickets that can be purchased
			// each goroutine MUST only buy 1 ticket.
			for i := 0; i < multiple; i++ {
				go buyTickets()
			}
```

vsp.go
```diff
@@ -57,65 +61,309 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	max := int(wal.Internal().ChainParams().MaxFreshStakePerBlock)
+	if buy > max {
+		buy = max
+	}
+
+	//TODO: limit should be set dynamically
+	// limit := 1
+	// if limit > 0 && buy > limit {
+	// 	buy = limit
+	// }
```
itswisdomagain, line 308: Restore this, but the limit should be passed as a parameter and should ALWAYS be 1 per my previous comment.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			buy := int(dcrutil.Amount(spendable) / sdiff)
+			if buy == 0 {
+				log.Debugf("Skipping purchase: low available balance on Wallet: %s", wal.Name)
+				return nil
+			}
+
+			max := int(wal.Internal().ChainParams().MaxFreshStakePerBlock)
+			if buy > max {
+				buy = max
+			}
```
itswisdomagain, line 264: Thought the goal was to buy as many as possible, no limits. Instead of reducing the quantity here, a log should do, advising the user that tb is about to purchase more than `max` tickets which means some tickets WILL NOT be included in the next block.

I believe the rationale behind this limiting in dcrwallet is to reduce the possibility of tickets that are purchased now, not making it into the next block. If you look above, you'll notice that ticket purchases are canceled if the next block will cause the ticket price to change and thus any tickets not already mined by then will be rejected (I think).

Probably worth discussing first though, just something to keep in mind for your next commit.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+
+			multiple := buy // multiple is the max amount of tickets that can be bought based on available funds
```
itswisdomagain, line 266: Do we need a new variable for this purpose? Can just keep using `buy` or rename `buy` to `multiple` above.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			buyTickets := func() {
+				err := vsp.buyTickets(cancelCtx, expiry, sdiff, passphrase)
+				if err != nil {
+					switch {
+					// silence these errors
+					case errors.Is(err, errors.InsufficientBalance):
+					case errors.Is(err, context.Canceled):
+					case errors.Is(err, context.DeadlineExceeded):
+					default:
+						log.Errorf("Ticket purchasing failed: %v", err)
```
itswisdomagain, line 279: Wallet ID/name.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			if buy > max {
+				buy = max
+			}
+
+			multiple := buy // multiple is the max amount of tickets that can be bought based on available funds
+
+			cancelCtx, cancel := context.WithCancel(ctx)
+			cancels = append(cancels, cancel)
+			buyTickets := func() {
+				err := vsp.buyTickets(cancelCtx, expiry, sdiff, passphrase)
```
itswisdomagain, line 271: dcrwallet uses this signature `cancelCtx, passphrase, tipHeader, expiry, &cfg`, shouldn't differ too much, this can be
```suggestion
				err := vsp.buyTickets(cancelCtx, passphrase, sdiff, expiry)
```

with `sdiff` replacing `tipHeader` for obvious reasons.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			// each purchase only buy 1 ticket.
+			for i := 0; i < multiple; i++ {
+				go buyTickets()
+			}
+		}
+	}
+}
+
+// buyTickets purchases one or more tickets depending on the spendable balance of the selected account,
+// the specified minimum balance to maintain and the limit per block.
```
itswisdomagain, line 300: Wrap comments into max 80 chars per line.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	wal := vsp.w
+
+	if len(passphrase) > 0 && wal.IsLocked() {
+		err := wal.UnlockWallet(passphrase)
+		if err != nil {
+			return translateError(err)
+		}
+	}
+
+	defer wal.LockWallet()
```
itswisdomagain, line 314: Lock only if you unlocked.
Also, multiple goroutines may use this method. If one complete before another and locks the wallet, it may affect the other purchase. Best to leave the wallet unlocked until tb is stopped.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+
+	request := &w.PurchaseTicketsRequest{
+		Count:                1,
```
itswisdomagain, line 317: It'd be wise to add a comment here, explaining why the `Count` is set to 1 always. To help future maintainers of this code understand the rationale behind the code and prevent unwanted changes.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	}
+
+	return err
+}
+
+// IsAutoTicketsPurchaseActive returns true if account mixer is active
+func (mw *MultiWallet) IsAutoTicketsPurchaseActive(walletID int) bool {
+	wallet := mw.WalletWithID(walletID)
+	if wallet == nil {
+		log.Errorf(ErrNotExist)
```
itswisdomagain, line 344: Don't think this error log is necessary, serves no purpose, at least not without a description.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+		return false
+	}
+
+	return wallet.cancelAutoTicketBuyer != nil
+}
+
+// StopAutoTicketsPurchase stops the active account mixer
```
itswisdomagain, line 351: Update comment for this method and the previous method.

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
 
-	v.vspClient = newVSPClient(vspHost, nil, sourceWallet.Internal())
-	vspInfo, err := v.GetInfo(ctx)
+	vsp.client = newVSPClient(vspHost, nil, sourceWallet.Internal())
+	vspInfo, err := vsp.GetInfo(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	v.ctx = ctx
```
itswisdomagain, line 69: `vsp.ctx` is no longer set, causes a crash when purchasing tickets from godcr #711.

itswisdomagain (changes requested):

vsp.go
```diff
@@ -57,65 +61,315 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			log.Errorf("Ticket buyer instance errored: %v", err)
+		}
+	}()
+
+	return nil
+}
+
+func (vsp *VSP) runTicketBuyer(ctx context.Context, balanceToMaintain int64, passphrase []byte) error {
+	if len(passphrase) > 0 && vsp.w.IsLocked() {
+		err := vsp.w.UnlockWallet(passphrase)
```
itswisdomagain, line 148: This method zeroes the passphrase, making it unusable below. Please update the `UnlockWallet` method to NOT alter the passphrase, that method has no way of knowing if it will be used subsequently. Callers ideally should be the ones to zero passphrases, when they're completely done with it.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	return nil
+}
+
+// StartTicketBuyer starts the automatic ticket buyer.
+func (vsp *VSP) StartTicketBuyer(balanceToMaintain int64, passphrase []byte) error {
+	if balanceToMaintain < 0 {
+		return errors.New("Negative balance to maintain given")
+	}
+
+	if vsp.w.cancelAutoTicketBuyer != nil {
```
itswisdomagain, line 123: Requires a read lock at least.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	vsp.w.cancelAutoTicketBuyerMux.Unlock()
+
+	go func() {
+		log.Infof("Running ticket buyer on Wallet: %s", vsp.w.Name)
+
+		err := vsp.runTicketBuyer(ctx, balanceToMaintain, passphrase)
+		if err != nil {
+			vsp.w.cancelAutoTicketBuyerMux.Lock()
+			vsp.w.cancelAutoTicketBuyer = nil
+			vsp.w.cancelAutoTicketBuyerMux.Unlock()
```
itswisdomagain, line 140: This should be cleared whether or not error is non-nil.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+
+			if ctx.Err() != nil {
+				log.Errorf("[%d] TicketBuyerV2 instance v1 canceled, account number: %s", vsp.w.ID, vsp.w.Name)
```
itswisdomagain, line 143: `TicketBuyerV2` and `instance v1`?

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			}
+			log.Errorf("[%d] Ticket buyer instance v1 errored: %v", vsp.w.ID, err)
```
itswisdomagain, line 145: Are there other versions? If no, the version info is not necessary.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+		}
+	}()
+
+	return nil
+}
+
+// runTicketBuyer executes the ticket buyer. If the private passphrase is incorrect,
+// or ever becomes incorrect due to a wallet passphrase change, Run exits with an
```
itswisdomagain, line 153:
```suggestion
// or ever becomes incorrect due to a wallet passphrase change, runTicketBuyer exits with an
```

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			buy := int(dcrutil.Amount(spendable) / sdiff)
+			if buy == 0 {
+				log.Debugf("Skipping purchase: low available balance on Wallet: %s", wal.Name)
+				return nil
+			}
+
+			max := int(wal.Internal().ChainParams().MaxFreshStakePerBlock)
+			if buy > max {
+				log.Infof("Ticket buyer is about to purchase %v tickets which is more the max ticket of %v per block.", buy, max)
+			}
```
itswisdomagain, line 271: Looks unnecessary now considering recent discussion around this.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			for i := 0; i < buy; i++ {
+				go buyTickets()
+			}
+		}
+	}
+}
+
+// buyTickets purchases one or more tickets depending on the spendable balance of
+// the selected account, the specified minimum balance to maintain and the limit
+// per block.
```
itswisdomagain, line 306: There's no more limit per block.
```suggestion
// the selected account and the specified minimum balance to maintain.
```

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+}
+
+// StopAutoTicketsPurchase stops the active ticket buyer
+func (mw *MultiWallet) StopAutoTicketsPurchase(walletID int) error {
+	wallet := mw.WalletWithID(walletID)
+	if wallet == nil {
+		return errors.New(ErrNotExist)
+	}
+
+	if wallet.cancelAutoTicketBuyer == nil {
```
itswisdomagain, line 356: Requires mutex read lock at least.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			log.Infof("Purchased ticket %v at stake difficulty %v from wallet: %s", hash, sdiff, wal.Name)
+		}
+	}
+
+	return err
+}
+
+// IsAutoTicketsPurchaseActive returns true if ticket buyer is active
+func (wallet *Wallet) IsAutoTicketsPurchaseActive() bool {
+	return wallet.cancelAutoTicketBuyer != nil
```
itswisdomagain, line 346: Mutex read lock.

vsp.go
```diff
@@ -57,65 +61,316 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	wallet := mw.WalletWithID(walletID)
+	if wallet == nil {
+		return errors.New(ErrNotExist)
+	}
+
+	if wallet.cancelAutoTicketBuyer == nil {
+		return errors.New(ErrInvalid)
+	}
+
+	wallet.cancelAutoTicketBuyer()
```
itswisdomagain, line 360: Requires lock, at least a read lock.

vsp.go
```diff
@@ -440,18 +695,187 @@ func (v *VSP) GetFeeAddress(ctx context.Context, ticketHash chainhash.Hash) (dcr
...
+	vspHost := wallet.ReadStringConfigValueForKey(TicketBuyerVSPHostConfigKey, "")
+
+	return &TicketBuyerConfig{
+		VspHost:           vspHost,
+		PurchaseAccount:   accNum,
+		BalanceToMaintain: btm,
+	}
+}
+
+// TicketBuyerConfigIsSet checks if ticket buyer config has been set for a selected wallet
```
itswisdomagain, line 833: This and a few other comments on this file are more than 80 chars long.

itswisdomagain (changes requested):
Some more cleanup.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	vsp.w.cancelAutoTicketBuyerMux.Lock()
+	vsp.w.cancelAutoTicketBuyer = cancel
+	vsp.w.cancelAutoTicketBuyerMux.Unlock()
+
+	if len(passphrase) > 0 && vsp.w.IsLocked() {
+		err := vsp.w.UnlockWallet(passphrase)
+		if err != nil {
+			return translateError(err)
+		}
+		defer vsp.w.LockWallet()
```
itswisdomagain, line 141: The wallet will be unlocked again as soon as the goroutine below starts (almost immediately after this lock). No point locking it here, if you ask me.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	vsp.w.cancelAutoTicketBuyerMux.Unlock()
+	if cancelFunc != nil {
+		return errors.New("Ticket buyer already running")
+	}
+
+	ctx, cancel := vsp.w.shutdownContextWithCancel()
+
+	vsp.w.cancelAutoTicketBuyerMux.Lock()
+	vsp.w.cancelAutoTicketBuyer = cancel
+	vsp.w.cancelAutoTicketBuyerMux.Unlock()
```
itswisdomagain, line 134: Simplify...
```suggestion
	vsp.w.cancelAutoTicketBuyerMux.Lock()
	if vsp.w.cancelAutoTicketBuyer != nil {
		vsp.w.cancelAutoTicketBuyerMux.Unlock()
		return errors.New("Ticket buyer already running")
	}

	ctx, cancel := vsp.w.shutdownContextWithCancel()
	vsp.w.cancelAutoTicketBuyer = cancel
	vsp.w.cancelAutoTicketBuyerMux.Unlock()
```

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	if len(passphrase) > 0 && vsp.w.IsLocked() {
+		err := vsp.w.UnlockWallet(passphrase)
+		if err != nil {
+			return translateError(err)
+		}
+		defer vsp.w.LockWallet()
+	}
+
+	go func() {
+		log.Infof("Running ticket buyer on Wallet: %s", vsp.w.Name)
```
itswisdomagain, line 145:
```suggestion
		log.Infof("Running ticket buyer on wallet: %s", vsp.w.Name)
```
Consider using consistent logging patterns for the ticket buyer. Other logging use `[%d]` and walletID.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+
+		err := vsp.runTicketBuyer(ctx, balanceToMaintain, passphrase)
+		if err != nil {
+			if ctx.Err() != nil {
+				log.Errorf("[%d] TicketBuyer instance canceled, account number: %s", vsp.w.ID, vsp.w.Name)
```
itswisdomagain, line 150: account ~number~ -> account name or just account.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			}
+			log.Errorf("[%d] Ticket buyer instance errored: %v", vsp.w.ID, err)
```
itswisdomagain, line 152: Should this be in an else block to avoid logging twice if the ctx is canceled?

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+		vsp.w.cancelAutoTicketBuyer = nil
+		vsp.w.cancelAutoTicketBuyerMux.Unlock()
+	}()
+
+	return nil
+}
+
+// runTicketBuyer executes the ticket buyer. If the private passphrase is incorrect,
+// or ever becomes incorrect due to a wallet passphrase change, runTicketBuyer exits with an
+// errors.Passphrase error.
```
itswisdomagain, line 165: First 2 lines are over the recommended 80 char per line limit.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			w := vsp.w.Internal()
+
+			// Don't perform any actions while transactions are not synced through
+			// the tip block.
+			rp, err := w.RescanPoint(ctx)
+			if err != nil {
+				return err
+			}
+			if rp != nil {
+				log.Debugf("Skipping autobuyer actions: transactions are not synced on Wallet: %s", vsp.w.Name)
```
itswisdomagain, line 210: For logging consistency, use wallet name or ID consistently. If using wallet name, no need to capitalize wallet:
```suggestion
				log.Debugf("Skipping autobuyer actions: transactions are not synced on wallet: %s", vsp.w.Name)
```

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+				intervalSize := int32(w.ChainParams().StakeDiffWindowSize)
+				currentInterval := height / intervalSize
+				nextIntervalStart = (currentInterval + 1) * intervalSize
+
+				// Skip this purchase when no more tickets may be purchased in the interval and
+				// the next sdiff is unknown.  The earliest any ticket may be mined is two
+				// blocks from now, with the next block containing the split transaction
+				// that the ticket purchase spends.
+				if height+2 == nextIntervalStart {
+					log.Debugf("Skipping purchase: next sdiff interval starts soon")
```
itswisdomagain, line 240: Wallet info in log.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			wal := vsp.w
+			// Get the account balance to determine how many tickets to buy
+			bal, err := wal.GetAccountBalance(int32(vsp.purchaseAccount))
+			if err != nil {
+				return err
+			}
+
+			spendable := bal.Spendable
+			if spendable < balanceToMaintain {
+				log.Debugf("Skipping purchase: low available balance on Wallet: %s", wal.Name)
```
itswisdomagain, line 263: ID or name consistently. Also `Wallet` -> `wallet`.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+
+			spendable -= balanceToMaintain
+			sdiff, err := wal.Internal().NextStakeDifficultyAfterHeader(ctx, tipHeader)
+			if err != nil {
+				return err
+			}
+
+			buy := int(dcrutil.Amount(spendable) / sdiff)
+			if buy == 0 {
+				log.Debugf("Skipping purchase: low available balance on Wallet: %s", wal.Name)
```
itswisdomagain, line 275: ID or name consistently. Also `Wallet` -> `wallet`.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+			// each purchase only buy 1 ticket.
+			for i := 0; i < buy; i++ {
+				go buyTickets()
+			}
+		}
+	}
+}
+
+// buyTickets purchases one or more tickets depending on the spendable balance of
+// the selected account and the specified minimum balance to maintain.
```
itswisdomagain, line 311: Comment appears to exceed the 80 char per line max. Also, no longer purchases more than one tickets.

vsp.go
```diff
@@ -57,65 +61,325 @@ func (mw *MultiWallet) NewVSPClient(vspHost string, walletID int, purchaseAccoun
...
+	cancelFunc := wallet.cancelAutoTicketBuyer
+	wallet.cancelAutoTicketBuyerMux.Unlock()
+	if cancelFunc == nil {
+		return errors.New(ErrInvalid)
+	}
+
+	wallet.cancelAutoTicketBuyerMux.Lock()
+	wallet.cancelAutoTicketBuyer()
+	wallet.cancelAutoTicketBuyer = nil
+	wallet.cancelAutoTicketBuyerMux.Unlock()
```
itswisdomagain, line 373: Simplify. Also, best to keep the lock until you're all done to avoid another process attempting to acquire the lock after you release it midway here.
```suggestion
	wallet.cancelAutoTicketBuyerMux.Lock()
	if wallet.cancelAutoTicketBuyer == nil {
		wallet.cancelAutoTicketBuyerMux.Unlock()	
		return errors.New(ErrInvalid)
	}
	
	wallet.cancelAutoTicketBuyer()
	wallet.cancelAutoTicketBuyer = nil
	wallet.cancelAutoTicketBuyerMux.Unlock()
```

vsp.go
```diff
@@ -440,18 +704,188 @@ func (v *VSP) GetFeeAddress(ctx context.Context, ticketHash chainhash.Hash) (dcr
...
+	vsp.client = newVSPClient(host, nil, sourceWallet.Internal())
+	vspInfo, err := vsp.GetInfo(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return vspInfo, nil
+}
+
+// SetAutoTicketsBuyerConfig set ticket buyer configuration for a particular wallet
```
itswisdomagain, line 822: 84 chars

vsp.go
```diff
@@ -440,18 +704,188 @@ func (v *VSP) GetFeeAddress(ctx context.Context, ticketHash chainhash.Hash) (dcr
...
+func (wallet *Wallet) SetAutoTicketsBuyerConfig(vspHost string, purchaseAccount int32, amountToMaintain int64) {
+	wallet.SetLongConfigValueForKey(TicketBuyerATMConfigKey, amountToMaintain)
+	wallet.SetInt32ConfigValueForKey(TicketBuyerAccountConfigKey, purchaseAccount)
+	wallet.SetStringConfigValueForKey(TicketBuyerVSPHostConfigKey, vspHost)
+}
+
+// GetAutoTicketsBuyerConfig gets ticekt buyer config for a selected wallet
+func (wallet *Wallet) GetAutoTicketsBuyerConfig() *TicketBuyerConfig {
```
itswisdomagain, line 830: `AutoTicketsBuyerConfig()` or just `TicketBuyerConfig()`, no need for the get prefix.

vsp.go
```diff
@@ -440,18 +704,188 @@ func (v *VSP) GetFeeAddress(ctx context.Context, ticketHash chainhash.Hash) (dcr
...
+	}
+}
+
+// TicketBuyerConfigIsSet checks if ticket buyer config has been set for a
+// selected wallet
+func (wallet *Wallet) TicketBuyerConfigIsSet() bool {
+	return wallet.ReadStringConfigValueForKey(TicketBuyerVSPHostConfigKey, "") != ""
+}
+
+// ClearTicketBuyerConfig clear all save ticket buyer config for a selected wallet
```
itswisdomagain, line 848: 83 chars. Also, save => saved.

wallet.go
```diff
@@ -39,6 +40,8 @@ type Wallet struct {
 	cancelFuncs        []context.CancelFunc
 	cancelAccountMixer context.CancelFunc
 
+	cancelAutoTicketBuyerMux sync.Mutex
```
itswisdomagain, line 43: Just seeing this properly, `mux` has a different meaning in go code I've seen.
```suggestion
	cancelAutoTicketBuyerMu sync.Mutex
```

itswisdomagain (changes requested):
A little more cleanup. I do have a cause for concern from testing but probably worthy of being investigated and fixed separately, most likely upstream even.

Where funds are available to purchase more than 1 ticket, two purchase attempts try to spend the same utxo leading to a double spend error. Ultimately, only one of the 2 ticket purchases succeed. The 2 log messages below show the 2 wallet methods I've seen to trigger this error.
```
2022-01-28 16:22:30.748 [ERR] DLWL: [1] Ticket purchasing failed: wallet.PurchaseTickets: double spend:: individualSplit:: wallet.processTransactionRecord: 4c535d6272423aec47eb8e1db22381484cd4b03202106af13ab852207ba55c00 conflicts with 7992bafc0203913c6ee32bcfc8b5c4c66dcdb5a4b4720269fcd11530814fa37d by double spending 9c70f0dbfbcec9d51b1506807f7514b23dd6e4572d106399c04361a3572fad71:0
2022-01-28 16:22:30.750 [ERR] DLWL: [1] Ticket purchasing failed: wallet.PurchaseTickets: double spend:: wallet.processTransactionRecord: f66453a8cd1cb5bb7383b98832f2756a994a36fb4d962dca032ea0b0ad546b6f conflicts with 7992bafc0203913c6ee32bcfc8b5c4c66dcdb5a4b4720269fcd11530814fa37d by double spending a6c65a88d836b0a149cde729fe1a2483d6e064a6e627139cea072ab2bc08b891:0
```

itswisdomagain (approved):

multiwallet_config.go
```diff
@@ -30,6 +30,11 @@ const (
 
 	VSPHostConfigKey = "vsp_host"
 
+	TicketBuyerVSPHostConfigKey = "tb_vsp_host"
+	TicketBuyerWalletConfigKey  = "tb_wallet_id"
```
itswisdomagain, line 34: Seems this isn’t used and can be removed.

dreacot (approved):

PR #216, opened 2021-12-09
add consensus voting capability 

dreacot:
Resolves https://github.com/planetdecred/godcr/issues/655

This diff adds consensus voting capability to this repo;

the capabilities present are;

- [x] get all agendas from the initial stake version till the current stake version
- [x] set vote choices for agendas on a per ticket basis
- [x] set vote choices for all tickets belonging to a wallet controlled by a VSP


consensus.go
```diff
@@ -0,0 +1,240 @@
...
+
+	return resp, nil
+}
+
+// SetVoteChoice handles a setvotechoice request by modifying the preferred
+// choice for a voting agenda.
+//
+// If a VSP host is configured in the application settings, the voting
+// preferences will also be set with the VSP.
+func (c *Consensus) SetVoteChoice(walletID int, vspPubKey []byte, vspHost, agendaID, choiceID, hash, passphrase string) error {
```
itswisdomagain, line 82: Comment needs some love.
`passphrase` should be `[]byte`.
`vspHost string`, before `vspPubKey []byte`.

consensus.go
```diff
@@ -0,0 +1,240 @@
...
+
+	ctx := wallet.shutdownContext()
+	_, err = wal.SetAgendaChoices(ctx, ticketHash, choice)
+	if err != nil {
+		return err
+	}
+
+	if vspHost == "" {
+		return fmt.Errorf("request requires a vspHost but no vspHost was provided")
+	}
```
itswisdomagain, line 117: If this is required, make this check much earlier. Also ensure that the vspClient can be created and return any error early.

consensus.go
```diff
@@ -0,0 +1,240 @@
...
+
+	vspClient, err := wallet.LookupVSP(vspHost)
```
itswisdomagain, line 119: Can do `wallet.VSP(host, pubkey)` directly, that will do a lookup first and create a new one if it doesn't already exist.

consensus.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+	}
+
+	// If ticket hash is provided, then set vote choice for the selected ticket
+	if ticketHash != nil {
+		err = vspClient.SetVoteChoice(ctx, ticketHash, choice)
+		return err
+	}
+
+	// Ticket hash wasn't provided, therefor set vote choice for all tickets
```
itswisdomagain, line 135:
```suggestion
	// Ticket hash wasn't provided, therefore set vote choice for all tickets
```

consensus.go
```diff
@@ -0,0 +1,240 @@
...
+	} else {
+		vspClient, err = wallet.VSP(vspHost, base64.StdEncoding.EncodeToString(vspPubKey))
+		if err != nil {
+			return err
+		}
+	}
+
+	// If ticket hash is provided, then set vote choice for the selected ticket
+	if ticketHash != nil {
+		err = vspClient.SetVoteChoice(ctx, ticketHash, choice)
```
itswisdomagain, line 131: If a ticket hash is provided, then it must be registered with this vsp. Is it confirmed that the vsp host provided has control of the ticket provided? Maybe better to get the actual, correct vsp host and pubkey from the wallet using `wal.VSPTicketInfo`. In that case, the caller shouldn't provide a host or pubkey, the code should get it from the wallet using the provided ticket hash.

go.sum
```diff
@@ -1,4 +1,6 @@
+bitbucket.org/creachadair/shell v0.0.6 h1:reJflDbKqnlnqb4Oo2pQ1/BqmY/eCWcNGHrIUO8qIzc=
```
itswisdomagain: This file shouldn't see changes without an accompanying change to go.mod. Fix with `go mod tidy`.

itswisdomagain (changes requested):
First pass.

consensus.go
dreacot: performing this check earlier prevents the  agenda choices from being saved locally, that should come first before the vsp attempt

consensus.go
dreacot: comment is same as upstream, but i can still modify it

consensus.go
dreacot: performing this check earlier prevents the  agenda choices from being saved locally, that should come first before the vsp attempt

consensus.go
```diff
@@ -0,0 +1,240 @@
...
+
+	choices, _, err := wal.AgendaChoices(ctx, ticketHash)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range choices {
+		resp.Choices[i] = VoteChoice{
+			AgendaID:          choices[i].AgendaID,
+			AgendaDescription: agendas[i].Vote.Description,
```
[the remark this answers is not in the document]
dreacot: Yeah, decrediton shows agendas from older versions as well, but the vote choices for that would default to abstain

consensus.go
dreacot: Screenshot from 2022-02-07 23-18-58

the `VSPTicketInfo` struct doesn't contain the pubkey

dreacot: Screenshot from 2022-02-07 23-18-58

the `VSPTicketInfo` struct doesn't contain the pubkey

consensus.go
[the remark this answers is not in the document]
itswisdomagain: More like will always be abstain unless decrediton has a way of getting the actual choices the user set then because the choices returned by `wallet.AgendaChoices` are only for the current vote version. Do we know if decrediton is able to fetch the actual vote choices for past agendas?

consensus.go
itswisdomagain: Yeah, the comment upstream makes more sense because there is an application setting that stores a pre-configured vsp host and also because the method is actually a handler for an rpc request. This one isn't an handler in that sense but does similarly.

consensus.go
itswisdomagain: Okay, so the vsp host isn't required. Perhaps this shouldn't return an error then? dcrwallet returns a nil error if there isn't a configured vsp host.

consensus.go
itswisdomagain: I think I meant `wal.Internal().VSPTicketInfo()`. That returns
```
type VSPTicket struct {
	FeeHash     chainhash.Hash
	FeeTxStatus uint32
	VSPHostID   uint32
	Host        string
	PubKey      []byte
}
```

consensus.go
dreacot: the vsp host is required, the method executes after agendas are saved locally, so it's still a requirement

consensus.go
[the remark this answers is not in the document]
dreacot: no indication currently of if decrediton is able to get the past vote choices

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	return resp, nil
+}
+
+func (mw *MultiWallet) agenda(walletID int, agendaID string) *chaincfg.ConsensusDeployment {
+	wallet := mw.WalletWithID(walletID)
+	wal := wallet.Internal()
+
+	_, agendas := getAllAgendas(wal.ChainParams())
```
itswisdomagain, line 67: The caller already has the `agendas`, not need to re-fetch them, let the caller pass in the slice. That way you also won't need the `walletID` parameter.

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+
+	for i := range choices {
+		agenda := mw.agenda(walletID, choices[i].AgendaID)
+		resp.Choices[i] = VoteChoice{
+			AgendaID:          choices[i].AgendaID,
+			AgendaDescription: agenda.Vote.Description,
+			ChoiceID:          choices[i].ChoiceID,
+			ChoiceDescription: "", // Set below
+		}
+		for j := range agendas[i].Vote.Choices {
```
itswisdomagain, line 52: `agenda.Vote.Choices` instead of `agendas[i].Vote.Choices`.
Also, isn't this better as `for _, choice := range `agenda.Vote.Choices`?

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	return nil
+}
+
+// SetVoteChoice sets a voting choice for an agenda using the set voting preference.
+// The voting preference is first saved locally, to the database.
+//
+// However, if a VSP host is configured in the application settings, the voting
+// preferences will also be set with the VSP.
```
itswisdomagain, line 82: Some lines are beyond the 80 char limit.
What do the following mean?
`sets a voting choice for an agenda using the set voting preference.`
`if a VSP host is configured in the application settings`.

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+func (mw *MultiWallet) SetVoteChoice(walletID int, vspHost string, vspPubKey []byte, agendaID, choiceID, hash string, passphrase []byte) error {
+	wallet := mw.WalletWithID(walletID)
+	if wallet == nil {
+		return fmt.Errorf("request requires a wallet but wallet has not loaded yet")
```
itswisdomagain, line 86: See previous comment on similar code.

consensus.go
itswisdomagain: If the vsp host is required, this method shouldn't do anything unless it is provided. It doesn't make much sense to save the choices locally, then return an error.

Also, if a ticket hash is provided, is the vsp host parameter also required? I expect that if a ticket hash is provided, then choices should be saved for only that ticket, not for any other ticket.

It looks to me that either a ticket hash or vsp host is required. Should validate early on that either is provided and valid, before saving the choice locally. Because an error response after saving the choice locally might give the user the impression that nothing was saved.

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+		vspClient, err = wallet.VSPClient(vspTicketInfo.Host, vspTicketInfo.PubKey)
+		if err != nil && !errors.Is(err, errors.NotExist) {
+			return err
+		}
+		err = vspClient.SetVoteChoice(ctx, ticketHash, choice)
+		return err
+	}
+
+	// Ticket hash wasn't provided, therefore set vote choice for all tickets
+	// belonging to the selected wallet and controlled by the set VSP
```
itswisdomagain, line 141:
```suggestion
	// belonging to the selected wallet and controlled by the specified VSP
```

itswisdomagain (changes requested):
Please document all public variables, types and methods.

consensus.go
dreacot: alright that makes sense, a check for the vsp host would be made earlier

yes that's how it's currently setup, if a ticket hash is provided the choices are saved for only that ticket

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+	if wallet == nil {
+		return fmt.Errorf("request requires a wallet but wallet has not loaded yet")
+	}
+	wal := wallet.Internal()
+
+	err := wallet.UnlockWallet(passphrase)
+	if err != nil {
+		return translateError(err)
+	}
+	defer wallet.LockWallet()
```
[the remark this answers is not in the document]
dreacot: we are testing if the passphrase is correct, because we need to ensure the voting is authorized, would add a comment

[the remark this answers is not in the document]
dreacot: we are testing if the passphrase is correct, because we need to ensure the voting is authorized, would add a comment to indicate

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+		return err
+	}
+
+	// If ticket hash is provided, then set vote choice for the selected ticket
+	if ticketHash != nil {
+		vspTicketInfo, err := wal.VSPTicketInfo(ctx, ticketHash)
+		if err != nil {
+			return err
+		}
+		// Register the provided ticket hash with the VSP client
```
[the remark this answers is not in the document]
dreacot: If voting for a particular ticket, the vspClient hasn't been initialized for that ticket which is what is being done here

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return err
+	}
+
+	if vspHost == "" {
+		return fmt.Errorf("request requires a vspHost but no vspHost was provided")
+	}
+
+	vspClient, err := wallet.VSPClient(vspHost, vspPubKey)
+	if err != nil && !errors.Is(err, errors.NotExist) {
```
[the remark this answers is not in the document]
dreacot: what if the user isn't voting for a particular ticket, but rather for all tickets controlled by a vsp, the client needs to be initialized with the provided vsphost

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+		vspTicketInfo, err := wal.VSPTicketInfo(ctx, ticketHash)
+		if err != nil {
+			return err
+		}
+		// Register the provided ticket hash with the VSP client
+		vspClient, err = wallet.VSPClient(vspTicketInfo.Host, vspTicketInfo.PubKey)
+		if err != nil && !errors.Is(err, errors.NotExist) {
+			return err
+		}
+		err = vspClient.SetVoteChoice(ctx, ticketHash, choice)
```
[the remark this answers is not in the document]
dreacot: not sure why i ignored that check, correcting it, thanks

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+		// get agenda status
+		currentTime := time.Now().Unix()
+		var status string
+		if currentTime > int64(d.ExpireTime) {
+			status = "Finished"
+		} else if currentTime > int64(d.StartTime) && currentTime < int64(d.ExpireTime) {
+			status = "In progress"
+		} else if currentTime > int64(d.StartTime) {
+			status = "Upcoming"
+		}
```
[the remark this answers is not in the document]
dreacot: to do this, i have to make the `type chaincfg.ConsensusDeployment` local, to define a new method on it, as the method we need to create is `ConsensusDeployment.Status()`

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+
+	ctx := wallet.shutdownContext()
+	choices, _, err := wal.AgendaChoices(ctx, ticketHash)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range choices {
+		agenda := mw.agenda(walletID, choices[i].AgendaID)
+		resp.Choices[i] = VoteChoice{
```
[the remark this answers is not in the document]
dreacot: that's correct, ultimately the missing values for the rest agendas doesn't break the code, they just aren't available

consensus.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		return nil
+	})
+
+	return firstErr
+}
+
+// GetAllAgendasForWallet returns all agendas through out the various stake versions for the active network and
+// this version of the software, and all agendas defined by it for a selected wallet.
+func (mw *MultiWallet) GetAllAgendasForWallet(walletID int, newestFirst bool) (*AgendasResponse, error) {
```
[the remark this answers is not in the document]
dreacot: `GetAllAgendasForWallet()` gets an agenda and all it's containing details for the specified wallet

`GetVoteChoices()` just gets only the vote choices for an agenda, they aren't similar

and we the major RPCs we are trying to get from upstream, are the `GetVoteChoices()` and `SetVoteChoice()`, it also makes it easier to maintain if we get changes upstream

They shouldn't be merged

types.go
```diff
@@ -493,3 +493,53 @@ type VSP struct {
...
+
+type Choice struct {
+	Id          string `json:"id"`
+	Description string `json:"description"`
+	Bits        uint32 `json:"bits"`
+	IsAbstain   bool   `json:"is_abstain"`
+	IsNo        bool   `json:"is_no"`
+}
+
+type GetVoteChoicesResult struct {
```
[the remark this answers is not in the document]
dreacot: yes that's true, how do you suggest we go about this, should we indicate multiple versions using an array?

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil, err
+	}
+
+	for i := range choices {
+		agenda := mw.agenda(agendas, choices[i].AgendaID)
+		resp.Choices[i] = AgendaVoteChoice{
+			AgendaID:          choices[i].AgendaID,
+			AgendaDescription: agenda.Vote.Description,
+			ChoiceID:          choices[i].ChoiceID,
+			ChoiceDescription: "", // Set below
```
Sirmorrison, line 50: Since this will be set below in same loop, there is no need setting it to `""` as its default values is `""`

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+// controlled by the set VSP.
+func (mw *MultiWallet) SetVoteChoice(walletID int, vspHost string, vspPubKey []byte, agendaID, choiceID, hash string, passphrase []byte) error {
+	wallet := mw.WalletWithID(walletID)
+	if wallet == nil {
+		return errors.New(ErrNotExist)
+	}
+	wal := wallet.Internal()
+
+	if vspHost == "" {
+		return fmt.Errorf("request requires a vspHost but no vspHost was provided")
```
Sirmorrison, line 86:
```suggestion
		return fmt.Errorf("request requires a vsp host but no vsp host was provided")
```

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+		return err
+	}
+
+	// If ticket hash is provided, then set vote choice for the selected ticket.
+	if ticketHash != nil {
+		vspTicketInfo, err := wal.VSPTicketInfo(ctx, ticketHash)
+		if err != nil {
+			return err
+		}
+		// Register the provided ticket hash with the VSP client.
```
Sirmorrison, line 127: you are not registering a ticket hash here... you are starting/initializing the vsp client. The registration is done on line 132.

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil
+	})
+
+	return firstErr
+}
+
+// GetAllAgendasForWallet returns all agendas through out the various
+// stake versions for the active network and this version of the software,
+// and all agendas defined by it for a selected wallet.
+func (mw *MultiWallet) GetAllAgendasForWallet(walletID int, newestFirst bool) (uint32, []*Agenda, error) {
```
Sirmorrison, line 155:
```suggestion
func (mw *MultiWallet) GetWalletAgenda(walletID int, newestFirst bool) (uint32, []*Agenda, error) {
```

Sirmorrison (changes requested):

consensus.go
dreacot: That's the way the code is done upstream, and i believe the reason for that is for better code readability, because at first glace if that variable isn't there and commented, the for loop might not be so obvious

consensus.go
dreacot: from below that comment, the steps to register a ticket hash is started, the main vsp client is registered on line 116.

To set the vote choices for a particular ticket hash, the vsp client needs to be reinitialized, using the vsp host and pub key for the selected ticket, that's why the comment starts from that line

consensus.go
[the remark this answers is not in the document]
itswisdomagain: Doesn't break the code but isn't correct either. Assuming there are 10 agendas in all, the first 3 won't be for the current vote version, but they will be assigned choices for the current vote version. I'd recommend fetching older agendas separately and the agendas for the current vote version separately, then set the choices for the latter ones. You're currently using data from one slice to update data in another slice but both slices are quite significantly different.

consensus.go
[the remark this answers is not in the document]
itswisdomagain: I think we're saying the same thing. If the user isn't voting for a specific ticket, then this initialization should happen a little lower in code, _after_ handling the case where the user is voting for a specific ticket.

consensus.go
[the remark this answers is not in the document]
itswisdomagain: > `GetAllAgendasForWallet()` gets an agenda and all it's containing details for the specified wallet

Gets an agenda? Or gets all agendas?

> `GetVoteChoices()` just gets only the vote choices for an agenda, they aren't similar

for an agenda? or for all agendas?
Also, doesn't `GetAllAgendasForWallet()` also get the vote choices for the agendas?

consensus.go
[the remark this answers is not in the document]
itswisdomagain: We can keep this for now, I suppose. A simple page refresh will likely cause this data to be reloaded and the current status updated.

types.go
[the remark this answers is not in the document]
itswisdomagain: Does the caller need the `version`?

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+
+	"decred.org/dcrwallet/v2/errors"
+	w "decred.org/dcrwallet/v2/wallet"
+	"github.com/decred/dcrd/chaincfg/chainhash"
+	"github.com/decred/dcrd/chaincfg/v3"
+)
+
+// GetVoteChoices returns configured vote preferences for each agenda of the
+// latest supported stake version for the wallet with the specified walletID.
+func (mw *MultiWallet) GetVoteChoices(hash string, walletID int) (*GetVoteChoicesResult, error) {
```
itswisdomagain, line 16: This isn't true though, it returns data for all agendas of all stake versions, not just the latest. Admittedly, the vote preferences are currently returned for only agendas of the latest stake version because there's currently no way to get previous agenda vote preferences. Comment needs an update to clarify the data being returned - data for all agendas of all stake versions but only preferences saved for the current stake version agendas.

To this end, I'm not sure this method should continue to be called `GetVoteChoices`, it doesn't really do that. The upstream method of this name only deals with the current stake version and thus is able to return the vote choices correctly, so it can correctly say it is getting vote choices. This seems to me to be getting all agendas (with vote choices for the latest/current agendas) rather than getting vote choices.

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+	if vspHost == "" {
+		return fmt.Errorf("request requires a vspHost but no vspHost was provided")
+	}
+
+	// Test to ensure the passphrase is correct, and the voting is authorized.
+	err := wallet.UnlockWallet(passphrase)
+	if err != nil {
+		return translateError(err)
+	}
+	defer wallet.LockWallet()
```
itswisdomagain, line 94: If we only unlocked to ensure that the passphrase is correct, we can lock right away. No need to defer the lock. If we need the wallet to remain unlocked, the comment above should say why.

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+	ctx := wallet.shutdownContext()
+	_, err = wal.SetAgendaChoices(ctx, ticketHash, choice)
+	if err != nil {
+		return err
+	}
+
+	vspClient, err := wallet.VSPClient(vspHost, vspPubKey)
+	if err != nil {
+		return err
+	}
```
itswisdomagain, line 119: vspClient is recreated in the if block below, this will never be used if ticketHash is not nil. It is only used if ticketHash is nil, after the if block below. This assignment should move to after the `if ticketHash != nil {` block.

consensus.go
itswisdomagain: What is being done here is setting the vote choice for the specified ticket with the vsp, NOT registering a ticket hash with the vsp client. Registering a ticket hash with a vsp client is done after the ticket is purchased and the fee is paid.

consensus.go
```diff
@@ -0,0 +1,235 @@
...
+		ticketHash = hash
+	}
+
+	choice := w.AgendaChoice{
+		AgendaID: agendaID,
+		ChoiceID: choiceID,
+	}
+
+	ctx := wallet.shutdownContext()
+	_, err = wal.SetAgendaChoices(ctx, ticketHash, choice)
```
itswisdomagain, line 111: Want to note that this may succeed but updating the vote choices with the vsp may fail. If that happens, the GetVoteChoices method will return this choice saved to the wallet as the current choice whereas that choice is only saved locally and is not sent to the vsp. That can be grossly misleading.

itswisdomagain (changes requested):
I'm still not very clear on the intended use of the new methods added in this PR.
- What purpose do the `mw.GetVoteChoices` and `mw.GetAllAgendasForWallet` serve? What UI process call those methods and why?
- Does SetVoteChoice really always require that a vsp host be provided? If a ticket hash is provided, shouldn't the method use the vsp host saved for that ticket hash rather than requesting the caller to provide a vsp host?

types.go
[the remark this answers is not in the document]
dreacot: No they don't

consensus.go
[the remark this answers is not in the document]
dreacot: for all agendas*

`GetAllAgendasForWallet()` makes a call to the get the `GetVoteChoices()` method, it doesn't directly get the vote choices for the agendas

dreacot:
@itswisdomagain based on your questions
Q1: Does SetVoteChoice really always require that a vsp host be provided? If a ticket hash is provided, shouldn't the method use the vsp host saved for that ticket hash rather than requesting the caller to provide a vsp host?

Ans: No, `SetVoteChoice()` doesn't always require a vsp host, the method uses the vsp host for the selected ticket (if provided), if not provided, it uses the user provided vsp host.

Q2: What purpose do the mw.GetVoteChoices and mw.GetAllAgendasForWallet serve? What UI process call those methods and why?

Ans: 
`GetVoteChoices()` simply returns the saved vote choices for the agendas, as per the RPC specificatin here https://github.com/decred/dcrwallet/blob/4f3b2d70ea259632fd8328fc010ddf30ec552aa2/internal/rpc/jsonrpc/methods.go#L122

`GetAllAgendasForWallet()` returns the agendas for the wallet with the provided ID, since agendas are on a per wallets basis, the user can select which wallet to get agendas for.

no UI process directly calls the `GetVoteChoices()` method as per design, while `GetAllAgendasForWallet()` is called by https://github.com/planetdecred/godcr/pull/729/files#diff-dec8e3ea6e8e79fc4373da512c8399413ebc858af71ca72a673e3739a2113e8aR133

consensus.go
dreacot: if voting with the vsp fails, the user would be notified with an error method, and i believe there should be a method that does reprocessing of the tickets, because saving the choices locally is what is done upstream as well, but if i'm wrong what do you suggest we do?

consensus.go
dreacot: noted

itswisdomagain:
> no UI process directly calls the `GetVoteChoices()` method as per design, while `GetAllAgendasForWallet()` is called by https://github.com/planetdecred/godcr/pull/729/files#diff-dec8e3ea6e8e79fc4373da512c8399413ebc858af71ca72a673e3739a2113e8aR133

If `GetVoteChoices` isn't required by any UI process currently, let's take it off. I know it's currently called by `GetAllAgendasForWallet()` but that's what even tipped me off as to its potential irrelevance. Modify `GetAllAgendasForWallet()` to do what needs to be done directly, without relying on `GetVoteChoices()`.

consensus.go
```diff
@@ -0,0 +1,236 @@
...
+	"decred.org/dcrwallet/v2/errors"
+	w "decred.org/dcrwallet/v2/wallet"
+	"github.com/decred/dcrd/chaincfg/chainhash"
+	"github.com/decred/dcrd/chaincfg/v3"
+)
+
+// GetVoteChoices returns configured vote preferences for each agenda of
+// all supported stake versions, but only preferences saved for the
+// current stake version agendas for the wallet with the specified walletID.
+func (mw *MultiWallet) GetVoteChoices(hash string, walletID int) (*GetVoteChoicesResult, error) {
```
itswisdomagain, line 17: Since this is wallet-specific, and does not really require access to the MultiWallet instance, I recommend making this and similar methods `(wallet *Wallet)` methods. Callers who do not have the desired wallet instance handy but have a `mw` instance and a walletID, can call `mw.WalletWithID` before calling the wallet method. I expect there'd be instances where the wallet instance is already available and can be used directly.

consensus.go
dreacot: Screenshot from 2022-02-28 11-42-17
Screenshot from 2022-02-28 11-42-01

this has been how our naming convention has been, so i believe `GetAllAgendasForWallet()`  is in line with that

consensus.go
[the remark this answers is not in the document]
dreacot: I have to disagree that's not what's happening on that line of code, at least not anymore, assuming there are 10 agendas in all, the first 3 would be assigned the vote choices for those particular agendas regardless of their position in the slice, the choices for those not in the current vote version are not even provided in that call, and are set as abstain at a later point in the code

consensus.go
dreacot: exactly that was my major concern, it deviating from upstream, right now i probably need to update my decrediton or so, as trying to vote on agendas doesn't work for me

but upstream might not have thought about this, so we don't necessarily need to copy them, we can revert the voting preference as a temp solution until we have better information?

consensus.go
```diff
@@ -0,0 +1,220 @@
...
+		if err != nil {
+			return 0, nil, fmt.Errorf("inavlid hash: %w", err)
+		}
+		ticketHash = hash
+	}
+
+	voteChoicesResult := &GetVoteChoicesResult{
+		Version: version,
+		Choices: make([]AgendaVoteChoice, len(deployments)),
+	}
```
itswisdomagain, line 126: Seems to me like you just only need a `choices := make([]AgendaVoteChoice, len(deployments))`. Maybe a different variable name.

consensus.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		for _, choice := range agenda.Vote.Choices {
+			if choices[i].ChoiceID == choice.Id {
+				voteChoicesResult.Choices[i].ChoiceDescription = choice.Description
+				break
+			}
+		}
+	}
+
+	for i := range deployments {
```
itswisdomagain, line 150: I'm quite convinced these 2 for loops can be merged. The first one only populates a temporary field that this one accesses.

types.go
```diff
@@ -493,3 +493,53 @@ type VSP struct {
...
+}
+
+// ByStartTime implements sort.Interface based on the Agenda.StartTime field.
+type ByStartTime []*Agenda
+
+func (a ByStartTime) Len() int { return len(a) }
+func (a ByStartTime) Less(i, j int) bool {
+	return a[i].StartTime > a[j].StartTime
+}
+func (a ByStartTime) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
```
itswisdomagain, line 543: A previous review recommended replacing this with a slice sorter that can be used e.g.:
```
sort.Slice(agendas, func(i, j int) bool {
	return a[i].StartTime > a[j].StartTime
})
```

utils.go
```diff
@@ -456,3 +457,18 @@ func TimeElapsed(now, then time.Time, abbreviationFormat string, fullTime bool)
 	}
 	return parts[0] + text
 }
+
+func voteVersion(params *chaincfg.Params) uint32 {
```
itswisdomagain, line 461: If callers don't currently care about the current vote version, we might be able to do without this function.

itswisdomagain (changes requested):
Looking better, but needs some more clean up still.

utils.go
dreacot: the caller doesn't require it, but it's still relevant in fetching all agendas, it's used here https://github.com/planetdecred/dcrlibwallet/pull/216/files#diff-51a9ecb87bb262c791b92cb917a6eca6d3f685e3bfce2638ddbce6c74fd115aeR206

we would need that method to be exported upstream, so we only have to maintain one copy of it

types.go
dreacot: oh, didn't come across that review

consensus.go
```diff
@@ -0,0 +1,245 @@
...
+		err = vspClient.SetVoteChoice(ctx, ticketHash, newChoice)
+		if err != nil {
+			// Updating the agenda voting preference with the vsp failed, therefore
+			// revert the locally saved voting preference for the agenda.
+			_, err = wal.SetAgendaChoices(ctx, ticketHash, currentChoice)
+			if err != nil {
+				return err
+			}
+		}
+
```
itswisdomagain, line 102: Ignore errors from trying to reset the prefs, return the original error.
```suggestion
		err = vspClient.SetVoteChoice(ctx, ticketHash, newChoice)
		if err != nil {
			// Updating the agenda voting preference with the vsp failed,
			// revert the locally saved voting preference for the agenda.
			_, revertError := wal.SetAgendaChoices(ctx, ticketHash, currentChoice)
			if revertError != nil {
				// log the revertError
			}
			return err
		}
```

consensus.go
```diff
@@ -0,0 +1,245 @@
...
+	// belonging to the selected wallet and controlled by the specififed VSP
+	var firstErr error
+	vspClient.ForUnspentUnexpiredTickets(ctx, func(hash *chainhash.Hash) error {
+		// Never return errors here, so all tickets are tried.
+		// The first error will be returned to the user.
+		err := vspClient.SetVoteChoice(ctx, hash, newChoice)
+		if err != nil && firstErr == nil {
+			// Updating the agenda voting preference with the vsp failed, therefore
+			// revert the locally saved voting preference for the agenda.
+			_, err = wal.SetAgendaChoices(ctx, ticketHash, currentChoice)
```
itswisdomagain, line 120: Don't revert here, this is a for loop. You might end up reverting more than once.
After the for loop completes, `if firstErr != nil`, then revert. Only log the revert error, but return firstError.

itswisdomagain (changes requested):

consensus.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/decred/dcrd/chaincfg/v3"
+)
+
+// SetVoteChoice sets a voting choice for the specified agenda. If a ticket
+// hash is provided, the voting choice is also updated with the VSP controlling
+// the ticket. If a ticket hash isn't provided, the vote choice is set for all
+// tickets controlled by the specified VSP.
+func (wallet *Wallet) SetVoteChoice(vspHost string, vspPubKey []byte, agendaID, choiceID, hash string, passphrase []byte) error {
+	wal := wallet.Internal()
+
```
itswisdomagain, line 18: Not sure this is needed anymore. Please just use `wallet.Internal()` directly instead. It's a little confusing seeing usages of `wal.***` and `wallet.***`.
```suggestion
```

consensus.go
```diff
@@ -0,0 +1,181 @@
...
+			break
+		}
+	}
+
+	newChoice := w.AgendaChoice{
+		AgendaID: agendaID,
+		ChoiceID: choiceID,
+	}
+
+	_, err = wal.SetAgendaChoices(ctx, ticketHash, newChoice)
```
itswisdomagain, line 65:
```suggestion
	_, err = wallet.Internal().SetAgendaChoices(ctx, ticketHash, newChoice)
```

consensus.go
```diff
@@ -0,0 +1,181 @@
...
+		// Update the vote choice for the ticket with the vsp
+		vspClient, err := wallet.VSPClient(vspTicketInfo.Host, vspTicketInfo.PubKey)
+		if err != nil {
+			return err
+		}
+		err = vspClient.SetVoteChoice(ctx, ticketHash, newChoice)
+		if err != nil {
+			// Updating the agenda voting preference with the vsp failed, therefore
+			// revert the locally saved voting preference for the agenda.
+			_, err = wal.SetAgendaChoices(ctx, ticketHash, currentChoice)
```
itswisdomagain, line 86: This will overwrite the initial error from `vspClient.SetVoteChoice`; we don't want that. Either just do `wallet.Internal().SetAgendaChoices(ctx, ticketHash, currentChoice)` without assigning the error OR assign the error to a different variable and log the error if non-nil.

consensus.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// Ticket hash wasn't provided, therefore set vote choice for all tickets
+	// belonging to the selected wallet and controlled by the specififed VSP
+	var firstErr error
+	vspClient.ForUnspentUnexpiredTickets(ctx, func(hash *chainhash.Hash) error {
+		// Never return errors here, so all tickets are tried.
+		// The first error will be returned to the user.
+		err := vspClient.SetVoteChoice(ctx, hash, newChoice)
+		if err != nil && firstErr == nil {
+
```
itswisdomagain, line 104:
```suggestion
```

consensus.go
```diff
@@ -0,0 +1,181 @@
...
+			firstErr = err
+		}
+		return nil
+	})
+
+	if firstErr != nil {
+		// Updating the agenda voting preference with the vsp failed, therefore
+		// revert the locally saved voting preference for the agenda.
+		_, err = wal.SetAgendaChoices(ctx, ticketHash, currentChoice)
+		log.Error(err)
```
itswisdomagain, line 114: Only log `if err. != nil`. Include a message as well.
```suggestion
		if err != nil {
			log.Errorf("unable to revert locally saved voting preference: %v", err)
		}
```

consensus.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return firstErr
+}
+
+// AllVoteAgendas returns all agendas through out the various
+// stake versions for the active network and this version of the software,
+// as well as returns the vote preferences for the current version,
+// and all agendas defined by it for a selected wallet.
```
itswisdomagain, line 123: This needs cleanup. Particularly the irregular characters-per-line length but also the last sentence looks off for me - don't understand what is meant there or how that sentence plays out with the rest of the comment.

I recommended:
```
// AllVoteAgendas returns all agendas of all stake versions for the active
// network and this version of the software. Also returns any saved vote
// preferences for the agendas of the current stake version. Vote preferences
// for older agendas cannot currently be retrieved.
```

utils.go
itswisdomagain, line 461: Please, add a comment here that this was borrowed from upstream and should be updated whenever the upstream version is also updated. The comment should also include a link to the LOC upstream where this function may be found.

itswisdomagain (changes requested):
We're getting there. After these changes, this should be good to go; but let's circle back to the godcr PR to be sure we won't need further changes here before merging this.

itswisdomagain (approved):
LGTM. Should be tested from https://github.com/planetdecred/godcr/pull/729 before merging.

consensus.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if newestFirst {
+		sort.Slice(agendas, func(i, j int) bool {
+			return agendas[i].StartTime > agendas[j].StartTime
+		})
+	}
+	return agendas, nil
+}
+
+func unmarshal(host string, target interface{}) error {
```
itswisdomagain, line 221: A similar function (`httpGet`) exists in the vsp.go file.

types.go
```diff
@@ -493,3 +495,36 @@ type VSP struct {
...
+	VotingPreference string            `json:"voting_preference"`
+	StartTime        int64             `json:"start_time"`
+	ExpireTime       int64             `json:"expire_time"`
+	Status           string            `json:"status"`
+}
+
+// AgendaTagged gets agenda information for the active network
+// from the dcrdata api https://dcrdata.decred.org/api/agendas for mainnet
+// or https://testnet.decred.org/api/agendas for testnet.
+type AgendaTagged struct {
```
itswisdomagain, line 516:
```suggestion
// DcrdataAgenda models agenda information for the active network from the
// dcrdata api https://dcrdata.decred.org/api/agendas for mainnet or
// https://testnet.decred.org/api/agendas for testnet.
type DcrdataAgenda struct {
```

wallet.go
```diff
@@ -289,3 +289,19 @@ func (wallet *Wallet) DecryptSeed(privatePassphrase []byte) (string, error) {
 
 	return decryptWalletSeed(privatePassphrase, wallet.EncryptedSeed)
 }
+
+// UnspentUnexpiredTickets returns all Unmined, Immature and Live tickets.
+func (wallet *Wallet) UnspentUnexpiredTickets() ([]Transaction, error) {
+	var tickets []Transaction
+	unspentUnexpiredTickets := []int32{TxFilterUnmined, TxFilterImmature, TxFilterLive}
+	for _, filter := range unspentUnexpiredTickets {
```
itswisdomagain, line 297:
```suggestion
	for _, filter := range []int32{TxFilterUnmined, TxFilterImmature, TxFilterLive} {
```

itswisdomagain (changes requested):

consensus.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/decred/dcrd/wire"
+)
+
+const (
+	DcrdataMainnetHost = "https://dcrdata.decred.org/api/agendas"
+	DcrdataTestnetHost = "https://testnet.decred.org/api/agendas"
+
+	AgendaStatusUpcoming   = "upcoming"
+	AgendaStatusInProgress = "in progress"
+	AgendaStatusFinished   = "finished"
```
itswisdomagain, line 24: Not sure these are used. Are they?

dreacot: https://github.com/planetdecred/godcr/pull/729/files#diff-dec8e3ea6e8e79fc4373da512c8399413ebc858af71ca72a673e3739a2113e8aR45 yeah, they are used

itswisdomagain (changes requested):
Brief, cursory look. Will re-review after reviewing the associated godcr PR.

consensus.go
```diff
@@ -0,0 +1,224 @@
...
+	AgendaStatusUpcoming = "upcoming"
+
+	// AgendaStatusInProgress used to define an agenda with voting ongoing,
+	// called from godcr.
+	AgendaStatusInProgress = "in progress"
+
+	// AgendaStatusFinished used to define an agenda that has finished voting,
+	// called from godcr.
+	AgendaStatusFinished = "finished"
+)
```
itswisdomagain, line 30: No need to add `called from godcr`. Other consumers of this public variables may find that confusing.

itswisdomagain (approved):
G2g after this.

PR #223, opened 2022-01-20
update spv pkg to match dcrwallet's v2.0.0-rc3 spv module

itswisdomagain:
Update spv pkg files to match those in https://github.com/decred/dcrwallet/tree/v2.0.0-rc3/spv.

The exception is the `(s *Syncer) DisableDiscoverAccounts()` added upstream for external consumers of the upstream pkg, for which we have no need currently. It can be added later if the need for it arises.

### DIFF
--- a/spv/backend.go
+++ b/spv/backend.go
@@ -55,7 +55,6 @@ func (wb *WalletBackend) Blocks(ctx context.Context, blockHashes []*chainhash.Ha
 
 // filterProof is an alias to the same anonymous struct as wallet package's
 // FilterProof struct.
-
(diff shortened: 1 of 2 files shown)


itswisdomagain:
Can manually compare dcrlibwallet's spv module and dcrwallet's using git:
```
git fetch https://github.com/decred/dcrwallet.git refs/tags/v2.0.0-rc3:refs/tags/dcrwallet2rc3
git diff dcrwallet2rc3 update-spv -- spv
```

Sirmorrison (approved):
LGTM

PR #226, opened 2022-01-24
get wallet deployement/activation block height

Sirmorrison:
This Pr gets the wallet activation/deployment block height.
This aim to prevent the godcr error `wallet.NextStakeDifficulty: inactive deployment: DCP0001 is not known to be active` when getting ticket price from a wallet when it has not reached the sync activation height on mainnet network.

### DIFF
--- a/sync.go
+++ b/sync.go
@@ -9,10 +9,12 @@ import (
 	"strings"
 	"sync"
 
+	"decred.org/dcrwallet/v2/deployments"
 	"decred.org/dcrwallet/v2/errors"
 	"decred.org/dcrwallet/v2/p2p"
 	w "decred.org/dcrwallet/v2/wallet"
 	"github.com/decred/dcrd/addrmgr/v2"
+	"github.com/decred/dcrd/chaincfg/v3"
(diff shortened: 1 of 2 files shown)


sync.go
```diff
@@ -491,3 +493,16 @@ func (mw *MultiWallet) GetLowestBlockTimestamp() int64 {
 	}
 	return timestamp
 }
+
+func (mw *MultiWallet) GetActivationBlockHeight() int32 {
```
itswisdomagain, line 497: - This should go to the multiwallet utils file.
- [Go methods should not be prefixed by `Get`](https://go.dev/doc/effective_go#Getters), just `ActivationBlockHeight` is recommended.
- Also, `ActivationBlockHeight` isn't descriptive, use `DCP0001ActivationBlockHeight`. Also add a comment that says `returns the hardcoded block height that the DCP0001 deployment activates at. DCP0001 specifies hard forking changes to the stake difficulty algorithm.`.

itswisdomagain (approved):

dreacot (approved):

PR #227, opened 2022-01-25
don't load wallets with missing db files on startup

itswisdomagain:
Such wallets are unusable and  would return an error when
a caller later tries to open all loaded wallets using
mw.OpenWallets.


Sirmorrison (approved):
LGTM

PR #234, opened 2022-02-25
update lib/pq to v1.10.4 build lib for android

JustinBeBoy:
When try to build lib for android, we have the problem:

`github.com/planetdecred/dcrlibwallet/vendor/github.com/lib/pq/user_posix.go:13:6: userCurrent redeclared in this block`

This issue occur in user_posix.go file and the lib/pq have new version update for fix it.

This PR update lib/pq to new version for fix it

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,4 +29,6 @@ require (
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 )
 
+replace github.com/lib/pq => github.com/lib/pq v1.10.4
+
 go 1.16
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,10 @@ require (
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 )
 
+// Older versions of github.com/lib/pq are required by politeia (v1.9.0)
+// and dcrdex (v1.10.3) but only v1.10.4 and above can be compiled for
+// the android OS using gomobile. This replace can be removed once any
+// of those projects update their github.com/lib/pq dependency.
 replace github.com/lib/pq => github.com/lib/pq v1.10.4
 
 go 1.16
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/dgraph-io/badger v1.6.2
 	github.com/jrick/logrotate v1.0.0
 	github.com/kevinburke/nacl v0.0.0-20190829012316-f3ed23dbd7f8
+	github.com/lib/pq v1.10.4 // indirect
```
itswisdomagain, line 24: Instead of an indirect require, use a replace at the end of this file so it can be removed when it's no longer needed. Also makes it easy to add a readable comment explaining why the replace is being done.
```
// Older versions of github.com/lib/pq are required by politeia (v1.9.0)
// and dcrdex (v1.10.3) but only v1.10.4 and above can be compiled for
// the android OS using gomobile. This replace can be removed once any
// of those projects update their github.com/lib/pq dependency.
replace github.com/lib/pq => github.com/lib/pq v1.10.4
```

Remember to run `go mod tidy` after editing the go.mod file.

Also, it'll be find to include a rewording of the comment suggested above in this PR description, to provide more information on why this PR is needed. The current description `This PR resolve the issue can not build lib for android` doesn't say why there's currently a build issue or what is done to fix the issue.

itswisdomagain (changes requested):
Tested and confirm this fixes the `gomobile bind -target=android` issue. Just one little improvement.

go.mod
```diff
@@ -29,4 +29,6 @@ require (
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 )
 
+replace github.com/lib/pq => github.com/lib/pq v1.10.4
```
itswisdomagain, line 32: Please add this comment before this replace, so code maintainers understand why this change was done; especially since `lib/pq` isn't a direct dependency.
```suggestion
// Older versions of github.com/lib/pq are required by politeia (v1.9.0)
// and dcrdex (v1.10.3) but only v1.10.4 and above can be compiled for
// the android OS using gomobile. This replace can be removed once any
// of those projects update their github.com/lib/pq dependency.
replace github.com/lib/pq => github.com/lib/pq v1.10.4
```

itswisdomagain (approved):

Sirmorrison (approved):

PR #241, opened 2022-04-04
NextAddress should return a different address from CurrentAddress

itswisdomagain:
Resolves https://github.com/planetdecred/dcrandroid/issues/483
In addition to incrementing the last returned address index, the
NextAddress method now also returns the address at the newly
incremented index.

This also ensures that CurrentAddress always returns the last requested
address.

### DIFF
--- a/address.go
+++ b/address.go
@@ -75,6 +75,9 @@ func (wallet *Wallet) AddressInfo(address string) (*AddressInfo, error) {
 	return addressInfo, nil
 }
 
+// CurrentAddress gets the most recently requested payment address from the
+// wallet. If that address has already been used to receive funds, the next
+// chained address is returned.
 func (wallet *Wallet) CurrentAddress(account int32) (string, error) {
 	if wallet.IsRestored && !wallet.HasDiscoveredAccounts {
 		return "", errors.E(ErrAddressDiscoveryNotDone)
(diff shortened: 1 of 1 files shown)


address.go
```diff
@@ -75,30 +75,40 @@ func (wallet *Wallet) AddressInfo(address string) (*AddressInfo, error) {
...
 func (wallet *Wallet) NextAddress(account int32) (string, error) {
 	if wallet.IsRestored && !wallet.HasDiscoveredAccounts {
 		return "", errors.E(ErrAddressDiscoveryNotDone)
 	}
 
-	addr, err := wallet.Internal().NewExternalAddress(wallet.shutdownContext(), uint32(account), w.WithGapPolicyWrap())
+	// NewExternalAddress increments the lastReturnedAddressIndex but does
+	// not return the address at the new index. That is returned below by
+	// CurrentAddress.
+	_, err := wallet.Internal().NewExternalAddress(wallet.shutdownContext(), uint32(account), w.WithGapPolicyWrap())
```
itswisdomagain, line 105: I decided against manually incrementing the last returned address index because of this gap policy. It would not be trivial to calculate the difference between the last used and last returned indices in order to force-wrap the returned index. I should add (in a forced commit shortly) that this is a work around until if and when upstream harmonizes `NewExternalAddress` and `CurrentAddress`.

dreacot (approved):
Looks good, and fixes the issue

PR #242, opened 2022-04-11
Multi: Remove duplicate txs when raw tx is returned for cases where a wallet is imported more than once.

crux25:
This PR removes duplicate transactions for multi wallet's GetTransactionsRaw calls when a wallet is imported more than once. This PR is part of [work on godcr](https://github.com/planetdecred/godcr/pull/881)

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -142,6 +142,17 @@ func (mw *MultiWallet) GetTransactionsRaw(offset, limit, txFilter int32, newestF
 		return transactions[i].Timestamp < transactions[j].Timestamp
 	})
 
+	// Remove duplicate transactions from wallets imported more than once.
+	txHashes := make(map[string]bool)
+	txs := make([]Transaction, 0)
+	for _, tx := range transactions {
+		if _, hasHash := txHashes[tx.Hash]; !hasHash {
+			txHashes[tx.Hash] = true
+			txs = append(txs, tx)
+		}
+	}
+	transactions = txs
+
 	if len(transactions) > int(limit) && limit > 0 {
 		transactions = transactions[:limit]
 	}


LasTshaMAN:
Just curious, is it possible to compare new wallet against existing wallets somehow and prevent duplicate wallet to be added to MultiWallet in the first place ? (if so, it seems like a better solution to me)

LasTshaMAN:
> Just curious, is it possible to compare new wallet against existing wallets somehow and prevent duplicate wallet to be added to MultiWallet in the first place ? (if so, it seems like a better/straightforward solution to me)

Oh well, I guess you provide some explanation here: https://github.com/planetdecred/godcr/pull/881, 

but from what I understand:
- `dcrlibwallet` will either have unencrypted seed passed to it by user (if a known wallet gets imported), or will have to delegate seed generation to `dcrwallet` (instead of delegating, generation code can be copy-pasted to `dcrlibwallet` so that it will have access to the seed)
- so we can track every seed in plain text in `dcrlibwallet`, but instead of plaintext a safer way would be to store and compare hashes of those seeds
- from what I understand encryption kicks in only after these stages I've described above, and keeps seed at max security when `dcrwallet` works with it

so, implementing it this way doesn't seem too challenging to me, am I missing something ?

Although, for existing users (who already have some wallets set up with `godcr` app) this will not work as intended ... but is it that big of a deal, given how early in development `godcr` is (if it is a big deal, maybe both features can be implemented ? because it seems the design I've outlined above provides better user experience anyway) ?

crux25:
> > Just curious, is it possible to compare new wallet against existing wallets somehow and prevent duplicate wallet to be added to MultiWallet in the first place ? (if so, it seems like a better/straightforward solution to me)
> 
> Oh well, I guess you provided some explanation here: [planetdecred/godcr#881](https://github.com/planetdecred/godcr/pull/881),
> 
> but from what I understand:
> 
>     * `dcrlibwallet` will either have unencrypted seed passed to it by user (if a known wallet gets imported), or will have to delegate seed generation to `dcrwallet` (instead of delegating, generation code can be copy-pasted to `dcrlibwallet` so that it will have access to the seed)
> 
>     * so we can track every seed in plain text in `dcrlibwallet`, but instead of plaintext a safer way would be to store and compare hashes of those seeds
> 
>     * from what I understand encryption kicks in only after these stages I've described above, and keeps seed at max security when `dcrwallet` works with it
> 
> 
> so, implementing it this way doesn't seem too challenging to me, am I missing something ?
> 
> Although, for existing users (who already have some wallets set up with `godcr` app) this will not work as intended ... but is it that big of a deal, given how early in development `godcr` is (if it is a big deal, maybe both features can be implemented ? because it seems the design I've outlined above provides better user experience anyway) ?

Having dcrlibwallet save and manage encrypted wallet seeds introduces an attack surface. So it ranks low in the solution hierarchy that we are looking to adopt. 
I am looking into the possibility of comparing the default account xpubs at the point when the wallet is imported.

Sirmorrison:
issue resolve by #244

PR #243, opened 2022-04-13
add agenda status types for failed and locked in agendas

dreacot:
Resolves https://github.com/planetdecred/godcr/issues/883

This PR accounts for the following agenda status types `defined`, `started`, `failed`, `lockedin`, `locked in` and `active`.
These statuses were not previously accounted for, which led to a crash in the client app


Sirmorrison (approved):
LGTM

PR #244, opened 2022-04-19
implement mw.WalletWithSeed method to identify already used seeds

itswisdomagain:
Since existing wallet seeds are not stored unecrypted and may not even be
stored at all if the user has backed up the seed, the check for already-used
seeds relies on comparing the account xpub of the default account of existing
wallets with the xpub of the same account using the new seed.


crux25:
The approach is solid and code functions according to specifications.

dreacot (approved):

PR #252, opened 2022-05-23
MultiWallet struct should be tread safe

LasTshaMAN:
(likely) Resolves [#865](https://github.com/planetdecred/godcr/issues/865)

This PR adds a mutexes around `MultiWallet.wallets` and `MultiWallet.badWallets` maps to prevent races during concurrent access.

Haven't tested these changes myself, though.


multiwallet.go
```diff
@@ -429,6 +434,60 @@ func (mw *MultiWallet) LinkExistingWallet(walletName, walletDataDir, originalPub
...
+	result := make(map[int]*Wallet, len(mw.wallets))
+
+	mw.walletsMu.RLock()
+	for key, value := range mw.wallets {
+		result[key] = value
+	}
+	mw.walletsMu.RUnlock()
+
+	return result
+}
```
LasTshaMAN, line 448: seems like copying this map is the simplest way to implement it (and maintain), likely performance shouldn't suffer much from it,

tried to rely on `find-and-replace` feature in my IDE (hopefully it didn't miss any occurrences)

multiwallet.go
```diff
@@ -429,6 +434,60 @@ func (mw *MultiWallet) LinkExistingWallet(walletName, walletDataDir, originalPub
...
+	mw.wallets[key] = newWallet
+	mw.walletsMu.Unlock()
+}
+
+// walletsUpdate is concurrently-safe way to delete from mw.wallets map.
+func (mw *MultiWallet) walletsDelete(key int) {
+	mw.walletsMu.Lock()
+	delete(mw.wallets, key)
+	mw.walletsMu.Unlock()
+}
```
LasTshaMAN, line 462: adding update/delete methods here instead of inlining lock-unlock mostly for conformance with `walletsReadCopy` func

dreacot:
@itswisdomagain please could you take a look this

multiwallet.go
```diff
@@ -275,7 +280,7 @@ func (mw *MultiWallet) OpenWallets(startupPassphrase []byte) error {
 		return err
 	}
 
-	for _, wallet := range mw.wallets {
+	for _, wallet := range mw.walletsReadCopy() {
```
itswisdomagain, line 283: Comment above applies here. Keep the mutex locked and prevent any other access to the map until the wallets are all opened.

multiwallet.go
```diff
@@ -286,11 +291,11 @@ func (mw *MultiWallet) OpenWallets(startupPassphrase []byte) error {
 }
 
 func (mw *MultiWallet) AllWalletsAreWatchOnly() (bool, error) {
-	if len(mw.wallets) == 0 {
+	if len(mw.walletsReadCopy()) == 0 {
```
itswisdomagain, line 294: Not sure there's an advantage to making and using a copy of the map here, the operation here doesn't take long at all. Should just do a lock and defer unlock at the top of this method and access the map directly.

multiwallet.go
```diff
@@ -429,6 +434,60 @@ func (mw *MultiWallet) LinkExistingWallet(walletName, walletDataDir, originalPub
 	})
 }
 
+// walletsReadCopy is concurrently-safe way to read from mw.wallets map.
+func (mw *MultiWallet) walletsReadCopy() map[int]*Wallet {
+	result := make(map[int]*Wallet, len(mw.wallets))
```
itswisdomagain, line 439: Lock before reading `len(mw.wallets)`. Also applies to the badWalletsReadyCopy method.

multiwallet.go
```diff
@@ -570,13 +629,13 @@ func (mw *MultiWallet) DeleteBadWallet(walletID int) error {
...
 	os.RemoveAll(wallet.dataDir)
-	delete(mw.badWallets, walletID)
+	mw.badWalletsDelete(walletID)
 
 	return nil
 }
 
 func (mw *MultiWallet) WalletWithID(walletID int) *Wallet {
-	if wallet, ok := mw.wallets[walletID]; ok {
+	if wallet, ok := mw.walletsReadCopy()[walletID]; ok {
```
itswisdomagain, line 638: No need to copy the entire map to access just one wallet, lock the mutex here, find and return the desired wallet, then unlock the mutex on return.

multiwallet.go
```diff
@@ -614,12 +673,12 @@ func (mw *MultiWallet) NumWalletsNeedingSeedBackup() int32 {
 }
 
 func (mw *MultiWallet) LoadedWalletsCount() int32 {
-	return int32(len(mw.wallets))
+	return int32(len(mw.walletsReadCopy()))
```
itswisdomagain, line 676: No need to make a copy of the map just to get the count of items. A simple lock/unlock would be more efficient here.

itswisdomagain (changes requested):
Will review again after this. As a rule of thumb, if an operation won't take long, don't make a copy of the wallets map or lock and unlock the mutex more than once.

LasTshaMAN:
yeah, like I've mentioned above my reasoning was "to find and fix the issue in all places, without missing any, and then we can optimize on top of that"

image

I'll address each comment soon

multiwallet.go
LasTshaMAN: missed this one, good catch

LasTshaMAN:
Cleaned up a bit and addressed all the comments above.

LasTshaMAN:
@itswisdomagain pls take a look when you can ^

PR #253, opened 2022-06-03
add function to set treasury policy

dreacot:
Resolves #https://github.com/planetdecred/godcr/issues/657

This PR introduces two new methods;

1. `SetTreasuryPolicy()` - This method saves the voting policy for treasury spends using the provided PiKey
2. `AllTreasuryPolicies()` - This method returns voting policies for treasury spends by a particular PiKey


treasury.go
```diff
@@ -0,0 +1,204 @@
...
+			return nil, fmt.Errorf("parameter contains invalid hexadecimal: %w", err)
+		}
+		var policy string
+		switch wallet.Internal().TreasuryKeyPolicy(pikey, ticketHash) {
+		case stake.TreasuryVoteYes:
+			policy = "yes"
+		case stake.TreasuryVoteNo:
+			policy = "no"
+		default:
+			policy = "abstain"
```
itswisdomagain, line 170: default is invalid, or maybe just empty string. "abstain" is not a valid treasury vote type. See `stake.CheckTreasuryVote`.

itswisdomagain (changes requested):

treasury.go
dreacot: this was implemented upstream https://github.com/decred/dcrwallet/blob/master/internal/rpc/jsonrpc/methods.go#L3760

dreacot: i can't seem to find any reference to `stake.CheckTreasuryVote` in dcrwallet, can you link me

treasury.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		res := []*TreasuryKeyPolicy{
+			{
+				PiKey:  PiKey,
+				Policy: policy,
+			},
+		}
+		if tixHash != "" {
+			res[0].TicketHash = tixHash
+		}
```
itswisdomagain, line 169: Same as...
```suggestion
		res := []*TreasuryKeyPolicy{
			{
				TicketHash: tixHash,
				PiKey:  PiKey,
				Policy: policy,
			},
		}
```

treasury.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// PiKeys returns the sanctioned Politeia keys for the current network.
+func (mw *MultiWallet) PiKeys() [][]byte {
+	return mw.chainParams.PiKeys
+}
+
+// PiKey returns a single politiea key (converted to string) from the
+// available politiea keys specified by the provided index.
+func (mw *MultiWallet) PiKey(index int) string {
```
itswisdomagain, line 202: This seems buggy, how would callers know which index to pass? Seems like a help function that should be added to godcr perhaps.

itswisdomagain (dismissed):
Looks good, just nits.

treasury.go
dreacot: knowing which index to call is a valid concern, maybe we can add the links to the available keys as part of the comment?

this method seems like it might be more useful in the library

thoughts?

treasury.go
```diff
@@ -0,0 +1,204 @@
...
+			continue // try next tHash
+		}
+
+		// Update the vote policy for the ticket with the associated VSP.
+		vspClient, err := wallet.VSPClient(vspTicketInfo.Host, vspTicketInfo.PubKey)
+		if err != nil && firstErr == nil {
+			firstErr = err
+			continue // try next tHash
+		}
+		err = vspClient.SetVoteChoice(ctx, tHash, nil, nil, policyMap)
```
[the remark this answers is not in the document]
dreacot: `treasuryPolicy` is the vote policy for a politeia key

`tspendPolicy` is the vote policy for a treasury spend transaction

treasury.go
itswisdomagain: There's a `mw` method that returns all the keys, the caller of this `mw.PiKey(index)` method should use that method instead and do...
```
key := hex.EncodeToString(mw.PiKeys()[index])
```

itswisdomagain (changes requested):
Mostly nits.

treasury.go
```diff
@@ -0,0 +1,194 @@
...
+// If a ticket hash is provided, the voting policy is also updated with the VSP
+// controlling the ticket. If a ticket hash isn't provided, the vote choice is
+// saved to the local wallet database and the VSPs controlling all unspent,
+// unexpired tickets are updated to use the specified vote policy.
+func (wallet *Wallet) SetTreasuryPolicy(PiKey, newVotingPolicy, tixHash string, passphrase []byte) error {
+	var ticketHash *chainhash.Hash
+	if tixHash != "" {
+		tixHash, err := chainhash.NewHashFromStr(tixHash)
+		if err != nil {
+			return fmt.Errorf("inavlid ticket hash: %w", err)
```
itswisdomagain, line 25:
```suggestion
			return fmt.Errorf("invalid ticket hash: %w", err)
```

treasury.go
```diff
@@ -0,0 +1,194 @@
...
+		ticketHash = tixHash
+	}
+
+	// The wallet will need to be unlocked to sign the API
+	// request(s) for setting this voting policy with the VSP.
+	err := wallet.UnlockWallet(passphrase)
+	if err != nil {
+		return translateError(err)
+	}
+	defer wallet.LockWallet()
```
itswisdomagain, line 36: Perform this unlocking after the pikey and newVotingPolicy checks. No need to unlock the wallet and immediately relock it if either the pikey or newVotingPolicy is invalid.

itswisdomagain (approved):
Last changes. G2g.

itswisdomagain (approved):

Sirmorrison (approved):
LGTG

PR #255, opened 2022-06-20
Backend sync services

### LINKED ISSUE #248: backend sync service
dcrlibwallet needs an explorer api backend that doesn't require spv, for instant usage with no sync wait.

crux25:
This PR provides the capability to fetch data from backend services. closes #248 

#### USAGE
The External Service is called from an instance of dcrlibwallet. For Example to get the best best block height from external service use: `multiwallet.ExternalService.GetBestBlock()`

#### Supported Function List

| Function                               | Description                                             |
| -------------------------------------- |---------------------------------------------------------|
| GetBestBlock()                         | `Fetch the best block height (int32).`                  |  
| GetBestBlockTimeStamp()                | `Best block time, as unix timestamp.`                   |
| GetCurrentAgendaStatus()               | `The current agenda and its status.`                    |
| GetAgendas()                           | `List of all agendas and their high level details`      |
| GetAgendaDetails(agendaId string)      | `The details for agenda whose ID is {agendaId}`         |
| GetTreasuryBalance()                   | `Fetch the current treasury balance`                    |
| GetTreasuryDetails()                   | `Get extra details about the treasury.`                 |
|                                        | `E.g balance, spent amount, added amount,tx `           |
| GetExchangeRate()                      | `Fetch exchange rate data summary`                      |
| GetExchanges()                         | `Fetch the current state data of all known exchanges`   |
| GetTicketFeeRateSummary()              | `The ticket fee rate summary`                           |
| GetTicketFeeRate()                     | `Top 25 ticket fees`                                    |
| GetNHighestTicketFeeRate(nHighest int) | `Get the {nHighest} ticket fees`                        |
| GetTicketDetails()                     | `All ticket details`                                    |
| GetNHighestTicketDetails(nHighest int) | `Get Top N {nHighest} ticket details`                   |
| GetAddress(address string)             | `returns the balances and transactions of an address`   |
| GetXpub(xPub string)                   | `Returns balances and transactions of an xpub`          |
| GetTicker(exch Exchange, market string)| `Returns market ticker data for supported exchanges.`   |
|                                        | `The market pair should be in the format: DCR-BTC or dcr-btc. e.g To fetch dcr-btc ticker information from Binance  use ExternalService.Getticker(dcrlibwallet.ext.Binance, "dcr-btc")`   |


api/types.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// ExchangeRates is the dcr and btc prices converted to fiat.
+	ExchangeRates struct {
+		BtcIndex  string               `json:"btcIndex"`
+		DcrPrice  float64              `json:"dcrPrice"`
+		BtcPrice  float64              `json:"btcPrice"`
+		Exchanges map[string]BaseState `json:"exchanges"`
+	}
+
+	ExchangeState struct {
```
LasTshaMAN, line 60: some structs are missing comments

api/service.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	err = json.Unmarshal(r, &addressState)
+	if err != nil {
+		return
+	}
+	return
+}
+
+//GetXpub Returns balances and transactions of an xpub
```
LasTshaMAN, line 287: //GetXpub -> // GetXpub

api/httpclient.go
```diff
@@ -0,0 +1,183 @@
...
+		done <- result{resp, err}
+	}()
+	// Wait for the read or timeout
+	select {
+	case r := <-done:
+		return r.resp, r.err
+	case <-timer.C:
+		errTimeout := errors.New("timeout reading from API")
+		return nil, errTimeout
+	}
```
LasTshaMAN, line 88: I think http request with timeout could be made using `context.WithTimeout` (and using this ctx with the actual request you are making), that will basically replace the need for `doTimeoutRequest` method

api/api_test.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		log.Error("Error creating chain params.")
+		return
+	}
+
+	service := NewService(chainParams)
+	height := service.GetBestBlock()
+	if reflect.TypeOf(height).Kind() != reflect.Int32 {
+		t.Errorf("api.GetBestBlock returned %v; want unit32", reflect.TypeOf(height))
+	}
```
LasTshaMAN, line 21: seems like this check will never fail, unless Golang devs introduce some bug in Go compiler or `reflect` package, so its kinda pointless for us to do here

api/api_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+func TestGetBestBlock(t *testing.T) {
+	chainParams, err := utils.ChainParams("mainnet")
+	if err != nil {
+		log.Error("Error creating chain params.")
+		return
+	}
+
+	service := NewService(chainParams)
+	height := service.GetBestBlock()
```
LasTshaMAN, line 18: seems like this test relies or "3rd party API" working while this test is executing, this will likely result in some intermittent failures of this test, and it could be annoying to go look at it every time it fails in false-positive manner (on CI especially),

I don't know what's the approach `dcrlibwallet` is following on this matter, just thought worth mentioning

api/service.go
```diff
@@ -0,0 +1,408 @@
...
+		conf.BaseUrl = mainnetBaseUrl
+	}
+
+	client := NewClient(conf)
+	client.ReqFilter = func(info RequestInfo) (req *http.Request, err error) {
+
+		req, err = http.NewRequest(info.Method, info.Url, bytes.NewBuffer([]byte(info.Payload.(string))))
+		if err != nil {
+			log.Error(err)
+			return
```
LasTshaMAN, line 61: probably better log errors once, at the very top instead of here (to keep it consistent)

api/httpclient.go
```diff
@@ -0,0 +1,183 @@
...
+	reqInfo := RequestInfo{
+		client:  c,
+		Method:  method,
+		Payload: payload,
+		Url:     rawurl,
+	}
+
+	if c.ReqFilter == nil {
+		err = errors.New("Request Filter was not set")
+		return
```
LasTshaMAN, line 111: could just return this error in one line

api/httpclient.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	req, err = c.ReqFilter(reqInfo)
+
+	if err != nil {
```
LasTshaMAN, line 116: usually, `err` should be handled on the very next line, no need to put a newline in between

api/service.go
```diff
@@ -0,0 +1,408 @@
...
+		Debug: true,
+	}
+	if chainParams.Name == chaincfg.TestNet3Params().Name {
+		conf.BaseUrl = testnetBaseUrl
+	} else {
+		conf.BaseUrl = mainnetBaseUrl
+	}
+
+	client := NewClient(conf)
+	client.ReqFilter = func(info RequestInfo) (req *http.Request, err error) {
```
LasTshaMAN, line 56: maybe ReqFilter -> ReqDecorator would be a more fitting name ?

api/service.go
```diff
@@ -0,0 +1,408 @@
...
+		AskPrice:       tempTicker.AskPrice,
+		BidPrice:       tempTicker.BidPrice,
+		LastTradePrice: tempTicker.LastPrice,
+	}
+
+	return
+}
+
+func (s *Service) getBittrexTicker(market string) (ticker *Ticker, err error) {
+	s.client.BaseUrl = bittrexBaseUrl
```
LasTshaMAN, line 366: same is missing in other similar methods

api/service.go
```diff
@@ -0,0 +1,408 @@
...
+			return ticker, errors.New("Bittrex doesn't support testnet")
+		}
+
+		return s.getBittrexTicker(market)
+	case KuCoin:
+		if s.chainParams.Name == chaincfg.MainNetParams().Name {
+			s.client.BaseUrl = kucoinBaseUrl
+		} else {
+			s.client.BaseUrl = KuCoinTestnetBaseUrl
+		}
```
LasTshaMAN, line 334: checking `chainParams` like that to set proper `BaseUrl` looks kinda messy to me, there gotta be a better way to do this,

and I assume this `Service` is meant to be used concurrently, meaning changing `BaseUrl` without proper go-routine synchronisation patterns is racy (and probably not what you want)

LasTshaMAN (commented):
Just was passing by, couple of things caught my eye, then couple more did ...

api/httpclient.go
```diff
@@ -0,0 +1,141 @@
...
+	"strings"
+	"time"
+)
+
+const (
+	// Default http client timeout in secs.
+	defaultHttpClientTimeout = 10 * time.Second
+)
+
+// Client is the base for http/https calls
```
dreacot, line 19: this comment should move down

api/service.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	err = json.Unmarshal(r, &xPubBalAndTxs)
+	if err != nil {
+		return
+	}
+	return
+}
+
+// GetTicker returns market ticker data for the supported exchanges.
```
dreacot, line 311: This comment can be made more elaborate, for example
1. specifying that Bittrex doesn't support testnet.
2. from the supported function list, it says `The market pair should be in the format: DRC-BTC.`, but for Binance i get  `error:400 Bad Request:'{"code":-1121,"msg":"Invalid symbol."}'`, perhaps the market formats for the diffrent exchanges can be shown
3. For i get `&{KuCoin DRC-BTC 0 0 0} <nil>`, using `DCR-BTC` market pair, the LastTradeRate is not returned

dreacot (changes requested):
while trying to call `GetExchangeRate()` & `GetExchanges()`

i get the following error 
```
<nil> 
error:503 Service Unavailable:'Exchange rate monitoring disabled.
'
```

when i try to call `GetXpub()`

i get the following error
```
<nil> 
error:500 Internal Server Error:'{"error":"Internal server error"}
'
```

api/service.go
dreacot: might be that DCR-BTC Isn't supported on kucoin

crux25: > might be that DCR-BTC Isn't supported on kucoin

DCR_BTC is supported on kucoin. You are running on testnet. They support other coins on testnet but they don't support dcr-btc.

crux25: > 

You are getting this errors because you are on testnet3. These exchanges doesn't support dcr-btc on testnet.

api/httpclient.go
```diff
@@ -0,0 +1,140 @@
...
+		Method  string
+		Url     string
+	}
+)
+
+// NewClient return a new HTTP client
+func NewClient(conf *ClientConf) (c *Client) {
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: true,
```
xaur, line 48: I'm not familiar with TLS in Go, but if it says "disable certificate checks", perhaps it's better to enable it by default to be safe, and explicitly disable in test code?

log.go
```diff
@@ -60,6 +61,7 @@ var (
 	cmgrLog      = backendLog.Logger("CMGR")
 	amgrLog      = backendLog.Logger("AMGR")
 	vspcLog      = backendLog.Logger("VSPC")
+	apiLog       = backendLog.Logger("API")
```
xaur, line 64: Random idea - I'd tag it `3PR` (3rd party requests) or `EXT` (external requests) or something like that to communicate that these are requests outside the P2P network.

api/service.go
```diff
@@ -0,0 +1,466 @@
...
+	if err != nil {
+		log.Error(err)
+		return -1
+	}
+	var blockDataBasic *BlockDataBasic
+	err = json.Unmarshal(r, &blockDataBasic)
+	if err != nil {
+		log.Error(err)
+		return -1
+	}
```
dmigwi, line 119: Its seems like this implementation is duplicated all through the methods.
I think creating a general function that is called by all of this duplicate implementations will reduce the code duplication. Consider having a such helper function:
``` golang
func (s *Service) queryExchange(exchange, method, url string, data interface) (error, interface) { 
          s.setBackend(exchange)
	  resp, err := s.client.Do(method, url, "")
	  if err != nil {
		return err, data
	}
	err = json.Unmarshal(resp, &data)
        // Its very important to first evaluate the data before returning the result.
        return err, data
	}
```
So now you can call `s.queryExchange(...)` wherever you see that whole is duplicated.

api/service.go
```diff
@@ -0,0 +1,466 @@
...
+func (s *Service) GetCurrentAgendaStatus() (agenda *chainjson.GetVoteInfoResult, err error) {
+	s.setBackend(DcrData)
+	r, err := s.client.Do("GET", "api/stake/vote/info", "")
+	if err != nil {
+		return
+	}
+	err = json.Unmarshal(r, &agenda)
+	if err != nil {
+		return
+	}
```
dmigwi, line 133: call the `s.queryExchange(...)` here and you will shorten all that code.

dmigwi (changes requested):
I have left quite a number of comments, let me know if any one of them is unclear.

plus your code seems to be deficient in comments. Would you kindly add some more code for the less tech savy guys in our team

api/httpclient.go
```diff
@@ -0,0 +1,135 @@
...
+	if req == nil {
+		return nil, errors.New("error: nil request")
+	}
+
+	c.dumpRequest(req)
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return response, err
+	}
+	c.dumpResponse(resp)
```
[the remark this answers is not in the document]
crux25: Doesn't log unless log level is set to debug. See comment above.

api/httpclient.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	dump, err := httputil.DumpRequest(r, true)
+	if err != nil {
+		log.Debug("dumpReq err: %v", err)
+	} else {
+		log.Debug("dumpReq ok: %v", string(dump))
+	}
+}
+
+func (c *Client) dumpResponse(r *http.Response) {
```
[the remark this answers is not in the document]
crux25: Doesn't log unless log level is set to debug. See comment above.

api/httpclient.go
```diff
@@ -0,0 +1,135 @@
...
+		responseStr := string(response)
+
+		res := "'" + responseStr + "'"
+		errStr = "\nerror:" + resp.Status + ":" + res
+		err = errors.New(errStr)
+	}
+	return response, err
+}
+
+func (c *Client) dumpRequest(r *http.Request) {
```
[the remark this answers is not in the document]
crux25: Doesn't log unless log level is set to debug. See comment above.

api/service.go
```diff
@@ -0,0 +1,466 @@
...
+	testnetXpubIdentifier    = "t"
+)
+
+var (
+	mainnetUrl = map[Backend]string{
+		Bittrex:   "https://api.bittrex.com/v3",
+		Binance:   "https://api.binance.com",
+		BlockBook: "https://blockbook.decred.org:9161/",
+		DcrData:   "https://mainnet.dcrdata.org/",
+		KuCoin:    "https://api.kucoin.com",
```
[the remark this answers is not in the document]
crux25: No, this information is supposed to be hard coded. 
These are links to exchange servers and don't serve any purpose in the config file except to confuse users, and it'll potentially break functionality if a user unintentionally modifies these links. If these links change, even power users find it difficult to hunt down replacement links. Best to have a maintainer take care of stuff like this. Same reason you seldom see update servers URLs in config files and why most crypto apps don't let you set the link to its exchange server.

api/service.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	testnetUrl = map[Backend]string{
+		Binance:   "https://testnet.binance.vision",
+		BlockBook: "https://blockbook.decred.org:19161/",
+		DcrData:   "https://testnet.dcrdata.org/",
+		KuCoin:    "https://openapi-sandbox.kucoin.com",
```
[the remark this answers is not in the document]
crux25: See comment above.

api/service.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	supportedBackends = []string{"Bittrex", "Binance", "BlockBook", "DcrData", "KuCoin"}
+)
+
+func NewService(chainParams *chaincfg.Params) *Service {
+	conf := &ClientConf{
+		Debug: true,
```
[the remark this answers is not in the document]
crux25: Good catch.

api/service.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	testnetUrl = map[Backend]string{
+		Binance:   "https://testnet.binance.vision",
+		BlockBook: "https://blockbook.decred.org:19161/",
+		DcrData:   "https://testnet.dcrdata.org/",
+		KuCoin:    "https://openapi-sandbox.kucoin.com",
+	}
+
+	supportedBackends = []string{"Bittrex", "Binance", "BlockBook", "DcrData", "KuCoin"}
```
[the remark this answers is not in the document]
crux25: Will require explicit conversion to string, where string values are used, and if required there will be no way to programmatically get a list of supported backends. I'll investigate your solution.

api/service.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+const (
+	Bittrex   Backend = "bittrex"
+	Binance   Backend = "binance"
+	BlockBook Backend = "blockbook"
+	DcrData   Backend = "dcrdata"
+	KuCoin    Backend = "kucoin"
+)
+
```
dmigwi, line 32: You could join this  const two declarations

api/service.go
```diff
@@ -0,0 +1,461 @@
...
+	if err != nil {
+		log.Error(err)
+		return -1
+	}
+
+	h, err := strconv.ParseInt(string(r), 10, 32)
+	if err != nil {
+		log.Error(err)
+		return -1
+	}
```
dmigwi, line 96: I think I made a comment about this duplicated code in the functions below...

api/httpclient.go
```diff
@@ -0,0 +1,135 @@
...
+	req, err = c.RequestFilter(reqInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	if req == nil {
+		return nil, errors.New("error: nil request")
+	}
+
+	c.dumpRequest(req)
```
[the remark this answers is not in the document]
dmigwi: Generally in an app that is ain't a server this excessive logging doesn't help.

api/service.go
dmigwi: This is the unique part that should be restricted in the specific function. Its actually possible to do it the way i suggested.

dmigwi: Otherwise common implementation should be moved to a collective function

dreacot (commented):
This looks good so far, we could be merging soon.

I'm interested in what we can do regarding this https://github.com/planetdecred/dcrlibwallet/pull/255#discussion_r947109438

dreacot (commented):
Every exported method and field should also be commented.

dmigwi (approved):

api/service.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// GetTicketDetails returns all ticket details see drcdata's MempoolTicketDetails for the spcific information
+// returned.
+func (s *Service) GetTicketDetails() (ticketDetails *apiTypes.MempoolTicketDetails, err error) {
+	reqConf := &ReqConfig{
+		method: http.MethodGet,
+		url:    "api/mempool/sstx/details",
+	}
+	tDetails := &apiTypes.MempoolTicketDetails{}
```
dreacot, line 238: i don't think there's any advantage in shortening this. `ticketDetails` is a better variable name, and it's not an unusually long name that we need to shorten it, i saw similar instances of this such as `aList` instead of `agendaList` etc. not a worthy trade off in my opinion.

It's not a blocker in getting this merged, just thought to mention

crux25: The shortened variable are temporary variables used only in the functions. You can see the longer versions in the method signature.

crux25: Good catch. I'll fix that.

multiwallet.go
```diff
@@ -104,6 +107,8 @@ func NewMultiWallet(rootDir, dbDriver, netType, politeiaHost string) (*MultiWall
 		txAndBlockNotificationListeners:  make(map[string]TxAndBlockNotificationListener),
 		accountMixerNotificationListener: make(map[string]AccountMixerNotificationListener),
 	}
+	// API service.
+	mw.ApiService = api.NewService(chainParams)
```
itswisdomagain, line 111: It's not immediately clear to me what `ApiService` mean. Can we find a more descriptive name for the package? `API` is a generic term, so is `Service`.

api/service.go
```diff
@@ -0,0 +1,391 @@
...
+	"strings"
+
+	"github.com/decred/dcrd/chaincfg/v3"
+	chainjson "github.com/decred/dcrd/rpc/jsonrpc/types/v3"
+	apiTypes "github.com/decred/dcrdata/v7/api/types"
+)
+
+type (
+	// Backend defines the backend type.
+	Backend string
```
itswisdomagain, line 17: `Backend` is a generic term. Saying `Backend` defines the backend type, doesn't help. Looking at the type name and comment, I still have no idea what sort of data this type is meant to store.

itswisdomagain: [Here's an example](https://github.com/decred/dcrwallet/blob/3b3e9e04b3c27123f8cb5bc3204924bbd0021e6c/wallet/network.go#L40-L46) of a custom type that is a backend, with sufficient commentary on what sort of backend it is.

```
// NetworkBackend provides wallets with Decred network functionality.  Some
// wallet operations require the wallet to be associated with a network backend
// to complete.
//
// NetworkBackend expands on the Peer interface to provide additional
// functionality for rescanning and filtering.
type NetworkBackend interface {
```

itswisdomagain (commented):
Some type and variable names giving me a bit of a head scratcher. Thoughts?

dreacot (approved):

PR #257, opened 2022-07-21
move dcr related functionlaity to it's own package

dreacot:
This PR moves all DCR related methods and data structure into it's own `dcr` package, assets are now located under the `wallets` package.


wallets/dcr/rescan.go
```diff
@@ -0,0 +1,141 @@
...
+			estimatedTotalRescanTime := int64(math.Round(float64(elapsedRescanTime) / rescanRate))
+			rescanProgressReport.RescanTimeRemaining = estimatedTotalRescanTime - elapsedRescanTime
+
+			rescanProgressReport.GeneralSyncProgress = &GeneralSyncProgress{
+				TotalSyncProgress:         rescanProgressReport.RescanProgress,
+				TotalTimeRemainingSeconds: rescanProgressReport.RescanTimeRemaining,
+			}
+
+			if wallet.blocksRescanProgressListener != nil {
+				wallet.blocksRescanProgressListener.OnBlocksRescanProgress(rescanProgressReport)
```
dmigwi, line 79: IMO dropping the word `rescan` from variable names would make the variables shorter and still retain a descriptive meaning.

wallets/dcr/rescan.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return nil
+}
+
+func (wallet *Wallet) CancelRescan() {
+	wallet.syncData.mu.Lock()
+	defer wallet.syncData.mu.Unlock()
+	if wallet.syncData.cancelRescan != nil {
+		wallet.syncData.cancelRescan()
+		wallet.syncData.cancelRescan = nil
```
dmigwi, line 127: This idea of a function and a variable sharing a name doesn't feel good.

wallets/dcr/sync.go
```diff
@@ -0,0 +1,484 @@
...
+	// init activeSyncData to be used to hold data used
+	// to calculate sync estimates only during sync
+	wallet.initActiveSyncData()
+
+	wallets := make(map[int]*w.Wallet)
+	wallets[0] = wallet.Internal()
+	wallet.WaitingForHeaders = true
+	wallet.Syncing = true
+
+	syncer := spv.NewSyncer(wallets, lp)
```
dmigwi, line 225: This looks better to me:
`syncer := spv.NewSyncer(map[int]*w.Wallet{0:wallet.Internal()}, lp)`

wallets/dcr/sync.go
```diff
@@ -0,0 +1,484 @@
...
+}
+
+func (wallet *Wallet) getBestBlock() int32 {
+	if wallet.Internal() == nil {
+		// This method is sometimes called after a wallet is deleted and causes crash.
+		log.Error("Attempting to read best block height without a loaded wallet.")
+		return 0
+	}
+
+	_, height := wallet.Internal().MainChainTip(wallet.ShutdownContext())
```
dmigwi, line 454: Does make two calls to `wallet.Internal()` make any difference here?

wallets/dcr/sync.go
```diff
@@ -0,0 +1,484 @@
...
+	if wallet.Internal() == nil {
+		// This method is sometimes called after a wallet is deleted and causes crash.
+		log.Error("Attempting to read best block timestamp without a loaded wallet.")
+		return 0
+	}
+
+	ctx := wallet.ShutdownContext()
+	_, height := wallet.Internal().MainChainTip(ctx)
+	identifier := w.NewBlockIdentifierFromHeight(height)
+	info, err := wallet.Internal().BlockInfo(ctx, identifier)
```
dmigwi, line 468: Here you've made 3 calls to `wallet.Internal()`

wallets/dcr/syncnotification.go
```diff
@@ -0,0 +1,678 @@
...
+	"golang.org/x/sync/errgroup"
+)
+
+func (w *Wallet) spvSyncNotificationCallbacks() *spv.Notifications {
+	return &spv.Notifications{
+		PeerConnected: func(peerCount int32, addr string) {
+			w.handlePeerCountUpdate(peerCount)
+		},
+		PeerDisconnected: func(peerCount int32, addr string) {
+			w.handlePeerCountUpdate(peerCount)
```
dmigwi, line 17: This is really confusing how `PeerConnected` and `PeerDisconnected` execute the same function with no changes to inputs. It be easier to just execute one since both of them might not return any difference.

dmigwi: reducing this unnecessary calls would reduce the number of mutex blocks requested especially on functions that require mutex blocks.

wallets/dcr/syncnotification.go
```diff
@@ -0,0 +1,678 @@
...
+
+	if w.syncData.activeSyncData.cfiltersFetchProgress.startCFiltersHeight == -1 {
+		w.syncData.activeSyncData.cfiltersFetchProgress.startCFiltersHeight = startCFiltersHeight
+	}
+
+	// wallet := w.WalletWithID(walletID)
+	w.syncData.activeSyncData.cfiltersFetchProgress.totalFetchedCFiltersCount += endCFiltersHeight - startCFiltersHeight
+
+	totalCFiltersToFetch := w.getBestBlock() - w.syncData.activeSyncData.cfiltersFetchProgress.startCFiltersHeight
+	// cfiltersLeftToFetch := totalCFiltersToFetch - w.syncData.activeSyncData.cfiltersFetchProgress.totalFetchedCFiltersCount
```
dmigwi, line 81: I presume this commented code is useless, why not delete it?

wallets/dcr/sync.go
```diff
@@ -0,0 +1,484 @@
...
+		beginFetchTimeStamp:      -1,
+		headersFetchTimeSpent:    -1,
+		totalFetchedHeadersCount: 0,
+	}
+
+	addressDiscoveryProgress := AddressDiscoveryProgressReport{
+		GeneralSyncProgress:       &GeneralSyncProgress{},
+		addressDiscoveryStartTime: -1,
+		totalDiscoveryTimeSpent:   -1,
+	}
```
dmigwi, line 88: Where addition is conducted later on this `-1`  default values should be compensated later on to avoid misrepresenting the actual figures.

wallets/dcr/syncnotification.go
dmigwi: Also if only you exported all those functions in `type Notification struct{}`, this whole implementation wouldn't be necessary that is unless there is something I am missing out on.

dmigwi (commented):
I am still new around and I am yet to understand the best practices recommended on this project. Below are my two cents after reviewing the code.
1. Most newly added functions don't  have descriptive comments on what the underlying code does. This becomes harder to evaluate if the specific function has all the necessary implementations without any extras.
2. Variable names are overly descriptive making them too long. This makes most simple implementations break the standard 80 characters a line thus the code is less legible and harder to identify errors/bugs.
3. Variables and functions chaining inside several structs feels excessive on most instances

dcr.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/planetdecred/dcrlibwallet/wallets/dcr"
+)
+
+func initializeDCRWallet(rootDir, dbDriver, netType string) (*storm.DB, string, error) {
+	var mwDB *storm.DB
+
+	rootDir = filepath.Join(rootDir, netType, "dcr")
+	err := os.MkdirAll(rootDir, os.ModePerm)
+	if err != nil {
+		return mwDB, "", errors.Errorf("failed to create dcr rootDir: %v", err)
```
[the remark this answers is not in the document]
dreacot: returning nil would give you an error in that instance which is why I initialized an empty instance at the top

wallets/dcr/rescan.go
```diff
@@ -0,0 +1,141 @@
...
+		wallet.syncData.mu.Lock()
+		wallet.syncData.rescanning = true
+		wallet.syncData.cancelRescan = cancel
+		wallet.syncData.mu.Unlock()
+
+		if wallet.blocksRescanProgressListener != nil {
+			wallet.blocksRescanProgressListener.OnBlocksRescanStarted(walletID)
+		}
+
+		progress := make(chan w.RescanProgress, 1)
```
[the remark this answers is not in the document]
dreacot: as of now, i'm not writing any new code, i just moved all the old code into a package, perhaps we can consider this during a code cleanup

wallets/dcr/rescan.go
```diff
@@ -0,0 +1,141 @@
...
+		go wallet.Internal().RescanProgressFromHeight(ctx, netBackend, startHeight, progress)
```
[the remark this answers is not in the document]
dreacot: as of now, i'm not writing any new code, i just moved all the old code into a package, perhaps we can consider this during a code cleanup

wallets/dcr/rescan.go
dreacot: as of now, i'm not writing any new code, i just moved all the old code into a package, perhaps we can consider this during a code cleanup

wallets/dcr/rescan.go
dreacot: as of now, i'm not writing any new code, i just moved all the old code into a package, perhaps we can consider this during a code cleanup

wallets/dcr/sync.go
dreacot: as of now, i'm not writing any new code, i just moved all the old code into a package, perhaps we can consider this during a code cleanup

wallets/dcr/sync.go
dreacot: same comments as above

wallets/dcr/syncnotification.go
dreacot: thanks for pointing out, should be uncommented

dmigwi (approved):
Since no major code changes happened, this can go in as it is!

Great work in getting this done this fast!

dcr.go
```diff
@@ -0,0 +1,55 @@
...
+	"decred.org/dcrwallet/v2/errors"
+
+	"github.com/asdine/storm"
+
+	bolt "go.etcd.io/bbolt"
+
+	"github.com/planetdecred/dcrlibwallet/wallets/dcr"
+)
+
+func initializeDCRWallet(rootDir, dbDriver, netType string) (*storm.DB, string, error) {
```
itswisdomagain, line 16: This doesn't really initialize a dcr wallet; the databases initialized here aren't specifically used for a single dcr wallet, one is used to store multiwallet data including ALL (dcr) wallets info and also general, non-wallet specific config data. This is really actually initializing the multiwallet. It also even initializes the log rotator which is not tied to any wallet in particular but is meant for all logging done by this library. I'd say let's move the body of this function back to the `NewMultiwallet` function, and also retain the rootDir without appending "dcr".

itswisdomagain: I suppose we'd need a function like this in the dcr package, perhaps a `InitializeWallet` function so it reads as `dcr.InitializeWallet`. I'm not yet completely sure what the `dcr.InitializeWallet` function would do but it would be quite similar to what we have here. It wouldn't/shouldn't initialize a new log rotator though, the caller of the function should provide a logger to use. It also likely won't need to init a politeia database since politeia data won't be stored per wallet.

multiwallet.go
```diff
@@ -6,46 +6,42 @@ import (
...
-	txAndBlockNotificationListeners map[string]TxAndBlockNotificationListener
+type MultiWallet struct {
+	DbDriver string
+	RootDir  string
+	DB       *storm.DB
 
-	blocksRescanProgressListener     BlocksRescanProgressListener
-	accountMixerNotificationListener map[string]AccountMixerNotificationListener
+	ChainParams *chaincfg.Params
+	Assets      *Assets
```
itswisdomagain, line 39: Do these properties NEED to be exported? They were private properties.

multiwallet.go
```diff
@@ -56,80 +52,62 @@ func NewMultiWallet(rootDir, dbDriver, netType, politeiaHost string) (*MultiWall
...
+		DB:          dcrDB,
+		ChainParams: chainParams,
+		Assets: &Assets{
+			DCR: struct {
+				Wallets     map[int]*dcr.Wallet
+				BadWallets  map[int]*dcr.Wallet
+				DBDriver    string
+				RootDir     string
+				DB          *storm.DB
+				ChainParams *chaincfg.Params
```
itswisdomagain, line 73: These aren't set or used yet. Can be removed till if and when they're needed.

multiwallet.go
```diff
@@ -56,80 +52,62 @@ func NewMultiWallet(rootDir, dbDriver, netType, politeiaHost string) (*MultiWall
...
-	// init database for saving/reading wallet objects
-	err = mwDB.Init(&Wallet{})
-	if err != nil {
-		log.Errorf("Error initializing wallets database: %s", err.Error())
-		return nil, err
-	}
-
-	// init database for saving/reading proposal objects
-	err = mwDB.Init(&Proposal{})
+	dcrDB, dcrRootDir, err := initializeDCRWallet(rootDir, dbDriver, netType)
```
itswisdomagain, line 55: Related to my first comment, this db isn't a dcrDB; it's a multiwallet db and should use the rootDir without appending `"dcr"`. This db would track all wallets for all assets but the individual wallet dbs would be in the asset subfolders.

multiwallet.go
```diff
@@ -56,80 +52,62 @@ func NewMultiWallet(rootDir, dbDriver, netType, politeiaHost string) (*MultiWall
...
 		} else {
-			mw.wallets[wallet.ID] = wallet
+			mw.Assets.DCR.Wallets[wallet.ID] = wallet
 		}
+
+		logLevel := wallet.ReadStringConfigValueForKey(LogLevelConfigKey, "")
+		SetLogLevels(logLevel)
+
+		// initialize Politeia.
+		wallet.NewPoliteia(politeiaHost)
```
itswisdomagain, line 110: wallet.NewPoliteia returns some values that currently ignored.
Also, politeia shouldn't be a wallet construct - it's not tied to any one wallet. It should be much like the `internal/vsp` package that is pretty much standalone but can work with any provided dcr wallet.

itswisdomagain (changes requested):
First pass; initial thoughts.

In addition to these, I've also not seen how to create a new dcr wallet or restore from seed and add the wallet to the multiwallet's list of dcr wallets. I was expecting to see something like `mw.AddDcrWallet` that creates or restores a dcr wallet and adds it to the multiwallet's list of wallets.

dcr.go
dreacot: appending "dcr" enables us keep all dcr database files under a single folder and seems easier to manage/track. I agree the method name is icky, and doesn't properly describe what's going on.

having separate rootdir for different assets looks advantageous in terms of folder structure
Screenshot from 2022-08-23 19-47-46

also do you think it's remotely possible we might want to be using different database drivers for different coins?

itswisdomagain: > having separate rootdir for different assets looks advantageous in terms of folder structure

I agree. What I was getting at is, the databases created here aren't even for all dcr wallets alone but for all wallets. It's the multiwallet parent database that tracks ALL wallets for all assets, which is used to populate/restore all the asset wallets when multiwallet is initialized. It also stores general config values that aren't related to any wallet.

The "dcr" subfolder should be reserved for dcr wallet databases, as in your screenshot. But there should also exist in the rootDir, a multiwallet.db file (currently called wallets.db, as in storing info for all wallets, which makes sense too).

> do you think it's remotely possible we might want to be using different database drivers for different coins?

Yes. Very possible.

dreacot:
> First pass; initial thoughts.
> 
> In addition to these, I've also not seen how to create a new dcr wallet or restore from seed and add the wallet to the multiwallet's list of dcr wallets. I was expecting to see something like `mw.AddDcrWallet` that creates or restores a dcr wallet and adds it to the multiwallet's list of wallets.

Previous implementation didn't make room for adding to the multiwallet wallets list

multiwallet.go
dreacot: that true, also i agree we shouldn't have politeia as a wallet struct, current problem is that politeia is now a wallet property, viewable when a dcr wallet is opened, and each dcr wallet can have it's own specific settings in regards to politeia.

but with the new internal politeia package that was merged, we'd have to take away this code anyways

multiwallet.go
dreacot: Screenshot from 2022-08-26 15-47-46

dreacot: db driver is not used, would remove that, the others are being used

multiwallet.go
dreacot: Screenshot from 2022-08-26 16-29-46

assets would need to be if we are going to access it's fields

